PR #1, opened 2019-06-30
Make license copyright generic

scgruber:
Copyright is jointly held between all contributors, the LICENSE file
should reflect that.


CollinValley:
Looks good to me

PR #3, opened 2019-07-07
Support Async Elements

CollinValley:
This is a rather large pull request. It creates a basic Async Element that can be strung together with Sync elements to for a basic serial chain of computation.  See the commit message and it's implementation for more.  It is still a bit rough, and I expect there to be further changes when we add support for reactors to the elements. But this is a good start.  I also included some changes to the Readme, again it is a little rough, but I wanted to get some foundations laid down.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 **/*.rs.bk
 Cargo.lock
 .idea
+*.code-workspace
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 # route-rs
+
 A multithreaded, modular software defined router library, written in rust.  Safe, fast, and extensible.
 
 Original click modular router: https://github.com/kohler/click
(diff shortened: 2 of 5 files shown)


ddowl (changes requested):
This commit seems distinct enough that it should be in a separate PR. Is that relatively easy to do?

ddowl (commented):
A few minor comments:

1. Check the spelling/spacing in the commit message.
2. It might be worth breaking up this commit into several: one that adds the `ForeverDrain`, one that renames `ElementLink`'s structure, and one that adds the spsc Async Element implementation.

Definitely go over the commit message, but you don't have to split up the commit if you don't feel like it.

CollinValley:
I think, for the sake of my own laziness, I'm going to leave these two together as one PR, in the future though I will separate them.

CollinValley:
So in the interest of my own laziness, I'm not going to break this up into multiple commits. This is mostly because we have a few lines that would have been changed in one commit, then changed again in the next commit, and I don't want to go through and rectify them.  I did clean up the commit messages for each, and also fixed a few spelling errors and grammar in the README.md.  At this point I am going to commit the pull request.  In the future, we will only take smaller commits that are better formatted,  especially as the codebase starts to grow beyond just three files.

PR #4, opened 2019-07-09
Cvalley working notify

CollinValley:
Two commits to change our codebase from using foreverDrain to the more proper task::current().notify(). See the commit messages for more verbosity


src/lib.rs
```diff
@@ -120,10 +120,10 @@ mod tests {
         let elem2_link = ElementLink::new(Box::new(elem1_link.backend), elem2);
         let elem3_link = AsyncElementLink::new(Box::new(elem2_link), elem3, default_channel_size);
 
-        let elem1_drain = ForeverDrain::new(0, Box::new(elem1_link.frontend));
-        let elem3_drain = ForeverDrain::new(1, Box::new(elem3_link.frontend));
+        let elem1_drain = ExhaustiveDrain::new(0, Box::new(elem1_link.frontend));
+        let elem3_drain = ExhaustiveDrain::new(1, Box::new(elem3_link.frontend));
 
-        let elem3_consumer = ForeverDrain::new(1, Box::new(elem3_link.backend));
+        let elem3_consumer = ExhaustiveDrain::new(2, Box::new(elem3_link.backend));
```
ddowl, line 131: Good catch on the drain ID.

src/lib.rs
```diff
@@ -82,7 +82,7 @@ mod tests {
     }
 
     #[test]
-    fn two_async_elements_no_waiting() {
+    fn two_async_elements_immediate_notify() {
```
ddowl: I see the point of changing the test name to signify the change in polling behavior, but that isn't what the `no_waiting` was originally specifying. `no_waiting` was referring to the packet generator input immediately yielding values (i.e. never yielding `Async::NotReady`) until it was empty (then sends `Async::Ready(None)`). To verify that the new `task::notify` behavior works correctly, let's construct a new test called `<element_arrangement>_source_packets_generated_on_interval` that uses a `LinearPacketGenerator` rather than `iter_ok`. 

It would be even nicer if we could specify `context`s like in RSpec so we could say

```
context "packets arrive every #{source_interval} seconds" do
  let packet_generator = LinearIntervalGenerator::new(time::Duration::from_millis(source_interval), 10);

  // ... all test examples that use the packet generator
end
```

I'm not sure this is possible with Rust's test framework though. If contexts don't exist, we'll just have to use long test names, which are only slightly more inconvenient.

src/lib.rs
```diff
@@ -106,7 +106,7 @@ mod tests {
     }
 
     #[test]
-    fn series_sync_and_async_no_waiting() {
+    fn series_sync_and_async_immediate_notify() {
```
ddowl: Same as above

src/lib.rs
CollinValley: Okay, I'll put in another commit to add the Linear Generator test, fixup the naming convention. Standby

ddowl (changes requested):
The change is pretty straightforward, but left a few comments on the tests.

I think it's great to have little asides in the commit messages explaining our thought process as we go about building. Someone trying to understand the code will probably reading through commit messages, and having helpful explanations of _why_ we made certain design decisions. I'd even take it a bit further and add in a link to any helpful documentation, if it exists.

There's also a run-on-sentence in the third paragraph.

src/lib.rs
CollinValley: Resolved in new, 3rd commit in the stack.

src/lib.rs
CollinValley: I had to push this as a new commit, since we remove the ForeverDrain in the 2nd commit and I did not want to fix the merge conflicts of squashing 1 and 3 together.  I'm not sure about making the tests have a bit more Rspec way, maybe someone else could look into that. I'm sure that the very least we could define common functions.

CollinValley:
Fixed the run-on sentence in the 1st commit, and dropped in some links that would be helpful to anyone auditing the code.

PR #5, opened 2019-07-11
Impl Future for AsyncElementLink.frontend

CollinValley:
Previously we implemented Stream for both the frontend and backend
portions of the AsyncElementLink. This makes some intuitive sense,
since a Stream is meant to stand for a continuous stream of futures,
which represents the sequence of packets being processed by an
element fairly well. However, Tokio's executor needs handles to Futures
rather than to Streams; we had to place a shim, such as the
ExhaustiveDrain utility element, in front of each stream before they are
provided to the runtime. In order to simplify the syntax, and reduce the
complexity of our API, I implemented Future for the frontend, so that
the frontend can be handed to the runtime directly.

In this implementation of Future, it never returns Async::Ready until
the element is ready to be torn down. An Element's frontend will continue
to process packets off its input stream in a loop until it can no longer
do work, in which case it will return Async::NotReady.

Note that processing logic has not actually changed, since this was the
functionality caused by placing the frontend stream into the
ExhaustiveDrain utility element.

Relevant tests were also adjusted, all tests pass after adjustment.

### DIFF
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -1,4 +1,4 @@
-use futures::{Stream, Async, Poll, task};
+use futures::{Future, Stream, Async, Poll, task};
 use crossbeam::crossbeam_channel::{bounded, Sender, Receiver, TryRecvError};
 
 pub type ElementStream<Input> = Box<dyn Stream<Item = Input, Error = ()> + Send>;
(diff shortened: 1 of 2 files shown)


ddowl (commented):
By changing `AsyncElementLink.frontend` to implement `Future` rather than `Stream`, we essentially are hardcoding a specific implementation of our "stream drain". In this case, the frontend will continue polling its input stream, putting packets onto its buffer channel for the backend to poll from, and notifying the backend task that it has work to do. You could argue that this implementation covers _all_ needs of the frontend, but I'm inclined to disagree. I imagine building out additional "middleware" functionality between the client's `process`ing packets, where you might be able to decorate our `*ElementLink` `Stream`s with additional functionality.

For example, we could add additional facilities for logging certain element's data to a file:
```
let element_link = ElementLink::new(input_stream, core_elem)
                              .withLogging(log_filename)
                              .withOtherMiddleware();
``` 
or if that's a little to weird to implement
```
let element_link = ElementLink::new(input_stream, core_elem)
let element_drain = LoggingDrain(element_link, log_filename);
```

Having a separate `Future` to consume a `AsyncElementLink.frontend` means we have the flexibility to add in different kinds of composable functionality, so I don't think this change is necessary atm.

CollinValley:
I think that the Stream that is actually wrapped in a weird Future before being handed to Tokio is a hack. 
It creates a lot of boilerplate, that clearly isn't needed.
It adds memory footprint, as well as an additional dynamic function call, which is expensive
It confuses the model, since the runtime is not directly talking to any elements that actually process packets.

I feel that any of the middlewares you could want would be better embedded inside the actual AsyncElementLink, it makes more sense with the model, you can get different Links based on which new function or builder syntax you use, and in the simple case nobody pays extra. 

I can let this simmer until we meetup in person, but I really think this change is a worthwhile one, though obviously I have my bias of being the author :).

CollinValley:
Per your last point we unfortunately can't have a Future and a Stream implemented and used concurrently on the same Frontend, since both of their polls need a &mut self.

CollinValley:
TEST THE WEBHOOK PLS IGNORE

CollinValley:
After some discussion with @ddowl and Sam Gruber in person we have decided to merge this PR, since it keeps things simpler.

PR #6, opened 2019-07-13
Convert AsyncElement to be truly asynchronous through proper use of task.notify() 

CollinValley:
This is a large commit, becuase quite a few things changed in order
to make this change work. In the future, I expect the commits to get
shorter as the codebase matures.

In this commit we:

Rename the Frontend and Backend members of AsyncElementLink to Consumer
and Provider, respectfully. This improves the clarity, and also avoids
the use of some fairly overloaded terms. The Consumer is responsible for
consuming the packets off the input stream to the AsyncElement and
processing them. The Provider, provides the output packets, which are
queued up in an internal channel, to a caller via its implementation of
`Stream`. Think of the Consumer as the input interface, and the Provider
as the output interface.

We now start to return Async::NotReady when we are not able to make forward
progress. At this point we are truly taking advantage of Tokio's runtime,
and our AsyncElement can now be run in parallel, concurrently,
cross-thread, cross-core, you name it. It is done in an efficient manner
so that we are yielding compute resources when no more work can be done.

This is enabled by passing `Task` handles from the Producer to the
Consumer, and vis versa, when either is unable to make progress. When
the other side has cleared whatever blocker caused the sleep in the
first place, it will awaken the slept `Task`.

There is some additional work done to ensure that during shutdown,
tasks are not left to sleep forever. This is best explained by the
comments surrounding the `Drop` and `Poll` function implementations.

### DIFF
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -66,79 +66,115 @@ pub trait AsyncElement {
     fn process(&mut self, packet: Self::Input) -> Self::Output;
 }
 
-pub struct AsyncElementFrontend<E: AsyncElement> {
-    input_stream: ElementStream<E::Input>,
-    channel: Sender<E::Output>,
-    element: E
-}
-
-pub struct AsyncElementBackend<E: AsyncElement> {
-    channel: Receiver<E::Output>
-}
-
+/// The AsyncElementLink is a wrapper to create and contain both sides of the
+/// link, the consumer, which intakes and processes packets, and the provider,
+/// which provides an interface where the next element retrieves the output
+/// packet.
 pub struct AsyncElementLink< E: AsyncElement> {
-    pub frontend: AsyncElementFrontend<E>,
-    pub backend: AsyncElementBackend<E>
-}
-
-impl<E: AsyncElement> AsyncElementFrontend<E> {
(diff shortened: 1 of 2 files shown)


src/api/mod.rs
```diff
@@ -66,101 +66,191 @@ pub trait AsyncElement {
...
-    channel: Receiver<E::Output>
-}
+impl<E: AsyncElement> AsyncElementLink<E> {
+    pub fn new(input_stream: ElementStream<E::Input>, element: E, queue_capacity: usize) -> Self {
 
-pub struct AsyncElementLink< E: AsyncElement> {
-    pub frontend: AsyncElementFrontend<E>,
-    pub backend: AsyncElementBackend<E>
-}
+        let (to_provider, from_consumer) = bounded::<Option<E::Output>>(queue_capacity);
```
scgruber, line 81: This is not a blocker, but I think `bounded` here is a difficult to read, and would be much clearer as `crossbeam_channel::bounded`

scgruber (approved):

ddowl (approved):

src/api/mod.rs
CollinValley: That is a good idea, it is a good idea to keep calls out to dependent libraries explicit, for clarity. Especially since bounded is already quite non-descriptive

PR #8, opened 2019-07-15
Add new test type, collected yield. 

CollinValley:
The commit messages should speak for themselves. But collected yield is a test where we actually collect the output from the router so that we can compare it to the expected output and produce a pass/fail case that is not determined solely by whether the runtime returns or not.  I imagine that at some point almost all of our tests will follow the collected yield model, but this is a good first step.  To support this I created a new generator and collector. I think they are suitably generic for now, we may need some more features down the road, but I don't know what that might be at this juncture. 

I also include a commit that reduces the verbosity of our test pieces, I figure we should really only be printing output on a test failure, so that the test output is not impossible to read. This might be contentious though so I welcome feedback on this move.

### DIFF
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -77,14 +77,15 @@ pub struct AsyncElementLink< E: AsyncElement> {
 
 impl<E: AsyncElement> AsyncElementLink<E> {
     pub fn new(input_stream: ElementStream<E::Input>, element: E, queue_capacity: usize) -> Self {
+        assert!( queue_capacity <= 1000, format!("Async Element queue_capacity: {} > 1000", queue_capacity));
 
         let (to_provider, from_consumer) = bounded::<Option<E::Output>>(queue_capacity);
-        let (await_provider, wake_provider) = bounded::<task::Task>(1);
-        let (await_consumer, wake_consumer) = bounded::<task::Task>(1);
+        let (await_consumer, wake_provider) = bounded::<task::Task>(1);
+        let (await_provider, wake_consumer) = bounded::<task::Task>(1);
 
         AsyncElementLink {
-            consumer: AsyncElementConsumer::new(input_stream, to_provider, element, await_consumer, wake_provider),
-            provider: AsyncElementProvider::new(from_consumer, await_provider, wake_consumer)
+            consumer: AsyncElementConsumer::new(input_stream, to_provider, element, await_provider, wake_provider),
+            provider: AsyncElementProvider::new(from_consumer, await_consumer, wake_consumer)
         }
     }
 }
@@ -167,10 +168,10 @@ impl<E: AsyncElement> Future for AsyncElementConsumer<E> {
             match input_packet_option {
                 None => {
                     return Ok(Async::Ready(()))
-                }
+                },
                 Some(input_packet) => {
                     let output_packet: E::Output = self.element.process(input_packet);
-                    if let Err(err) = self.to_provider.send(Some(output_packet)) {
+                    if let Err(err) = self.to_provider.try_send(Some(output_packet)) {
                         panic!("Error in to_provider sender, have nowhere to put packet: {:?}", err);
                     }
                     if let Ok(task) = self.wake_provider.try_recv() {
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -1,5 +1,6 @@
 use futures::{Future, Stream, Async, Poll, task};
-use crossbeam::crossbeam_channel::{bounded, Sender, Receiver, TryRecvError};
+use crossbeam::crossbeam_channel::{Sender, Receiver, TryRecvError};
+use crossbeam::crossbeam_channel;
 
 pub type ElementStream<Input> = Box<dyn Stream<Item = Input, Error = ()> + Send>;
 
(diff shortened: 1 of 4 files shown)


src/utils/test/packet_collectors.rs
```diff
@@ -33,3 +34,39 @@ impl<T: Debug> Future for ExhaustiveDrain<T> {
...
+/// receives a None, but it also will write that value out to the provided channel, so that the packet
+/// may be compared in a test.
+pub struct ExhaustiveCollector<T: Debug> {
+    id: usize,
+    stream: ElementStream<T>,
+    packet_dump: crossbeam_channel::Sender<T>
+}
+
+impl<T: Debug> ExhaustiveCollector<T> {
+    pub fn new(id: usize, stream: ElementStream<T>, packet_dump: crossbeam_channel::Sender<T>) -> Self {
```
ddowl: In other places we've been typing by just `Sender<T>`. Could we import the whole `crossbeam::crossbeam_channel::Sender` please?

src/utils/test/packet_generators.rs
```diff
@@ -52,3 +52,47 @@ impl Stream for LinearIntervalGenerator {
         }
     }
 }
+
+/// Packet Interval Generator procduces a Stream of packets on a defined interval.AsMut
```
ddowl, line 56: The "LineraInterval" of `LinearIntervalGenerator` was meant to indicate that some packets are generated at a _fixed_ interval, e.g. every 100ms. The name `PacketIntervalGenerator` doesn't really make sense in this way. Maybe we should rename it to `FixedIntervalGenerator`, since you could also interpret the "Interval" as the ranges of digits, but that's not the intention. Let's deprecate `LinearIntervalGenerator`, since all ranges can be expressed by passing `range.into_iter()` to a `FixedIntervalGenerator`.

ddowl: We should also try to deprecate `LinearIntervalGenerator` since there's a lot of duplicated logic between the two.

src/utils/test/packet_collectors.rs
```diff
@@ -33,3 +34,39 @@ impl<T: Debug> Future for ExhaustiveDrain<T> {
...
+        ExhaustiveCollector { id, stream, packet_dump }
+    }
+}
+
+impl<T: Debug> Future for ExhaustiveCollector<T> {
```
ddowl, line 49: I wish there was a way to reduce redundancy, since `ExhaustiveDrain` and `ExhaustiveCollector` are so similar.

Could `ExhaustiveDrain` just contain an `ExhaustiveCollector` that does nothing with its channel?

src/lib.rs
```diff
@@ -52,6 +52,26 @@ mod tests {
         tokio::run(consumer);
     }
 
+    #[test]
+    fn one_sync_element_collected_yield() {
```
ddowl, line 55: I'm unsure why we're only adding more tests here instead of replacing all of the existing tests with the packet collection assertion. Why?

Also, when all of our tests pass/fail on assertions, the `collected_yield` suffix will be redundant.

src/lib.rs
```diff
@@ -25,7 +26,7 @@ mod tests {
         type Output = i32;
 
         fn process(&mut self, packet: Self::Input) -> Self::Output {
-            println!("Got packet {} in element {}", packet, self.id);
+            //println!("Got packet {} in element {}", packet, self.id);
```
ddowl: In any case, we shouldn't be committing commented code, but I don't think removing logging altogether is a good idea. It would be better if we could enable logging with via an environment variable, at least until we can enable `tokio_trace`.

I think something like [env_logger](https://docs.rs/env_logger/0.6.2/env_logger/) does the job quite well. That can be introduced in a separate commit. I'd be happy to look into that if you want.

src/api/mod.rs
```diff
@@ -77,14 +77,15 @@ pub struct AsyncElementLink< E: AsyncElement> {
 
 impl<E: AsyncElement> AsyncElementLink<E> {
     pub fn new(input_stream: ElementStream<E::Input>, element: E, queue_capacity: usize) -> Self {
+        assert!( queue_capacity <= 1000, format!("Async Element queue_capacity: {} > 1000", queue_capacity));
```
ddowl: I think this commit is useful, but should it be in this PR?

ddowl (changes requested):
I like the main ideas of the generic generator and packet collector here. There are a few comments that we should address before merging.

src/lib.rs
CollinValley: Yes, I think this is true. Can we clean this up later during the testing fixup. I was did not want to depreciate existing work until this PR is in.

src/lib.rs
CollinValley: Yes that would be a good idea, leave in for now,  resolve in new commit.

src/utils/test/packet_collectors.rs
CollinValley: Let's leave this for a separate PR.  We are going to drop many tests all together.

src/utils/test/packet_generators.rs
CollinValley: Leave deprecation for testing fixup issue.  I get the names are bad, but we shouldn't change too much in one PR

src/utils/test/packet_collectors.rs
CollinValley: Will clean this up.  I'm going to go with calls to `bounded` will be `crossbeam_channel::bounded`,  calls to Sender and Receiver will be `Sender` and `Receiver`.   Similar pattern for unbounded.  I think that balances readability with succinctness.

src/api/mod.rs
CollinValley: Hmm, that was probably added slightly by mistake.  Will fork that out somewhere else if at all possible.

CollinValley: Removed this to be added back in on a later PR

src/utils/test/packet_collectors.rs
CollinValley: Cleaned up as requested. Future PRs will also follow this convention.

ddowl (approved):

PR #20, opened 2019-07-20
add packet structs compatible with zerocopy trait

moosingin3space:


ddowl (changes requested):
Can you describe why we have a `PacketOwned` and a `PacketRef`? Collin and I have been leaving descriptive commit messages, like in the Meraki style. You should make your commit message be descriptive enough that anyone working on the project could figure out why we made certain decisions just from reading it. The code change itself looks good to me though!

CollinValley:
Echo Drew's concern, we should keep the commit messages fairly verbose so that someone auditing the code can use git blame to figure out why things are where they are.

moosingin3space:
Closing because I think it'll be easier to develop this with netif in place first.

PR #22, opened 2019-07-24
Draft: The MEG [ Test rework, changing api code layout, implement ClassifyElement, implement JoinElement, Add Overseer]

CollinValley:
I know this is kind of a mega commit stack, I just wanted you all to take a look at it before we try to open up individual PRs.  This solves quite a few of our open issue, and I think after this is merged we can start trying to write elements, and form there create a functioning router. Which is pretty cool.  As far as I am aware, this code works as written. We could use more tests as always, but it is a good start.


scgruber:
This is probably also a good opportunity to drop everything down a level so we can have multiple crates in the repo.

CollinValley:
Moving the api to it's own crate seems like a good idea.  I can do that before pushing the official PRs.

CollinValley:
Heads up: Resolved the Join element fairness issue this morning, commit is now in stack.

ddowl:
We can get those first 2 commits, moving the `Element` definitions to their own modules (`086f708`, `0930c69`), into a PR straight away and get them merged. Those lgtm.

ddowl:
I'm glad we're cleaning up the tests to be assertion-only. We should make a similar test cleanup commit for the synchronous element as well.

src/api/classify_element.rs
```diff
@@ -0,0 +1,349 @@
...
+            }
+        }
+    }
+}
+
+impl<E: ClassifyElement> Future for ClassifyElementConsumer<E> {
+    type Item = ();
+    type Error = ();
+
+    /// Same logic as AsyncElementConsumer, except if any of the channels are full we
```
ddowl, line 168: We can do this in a separate PR, but if we're sharing similar logic with `AsyncElementConsumer`, we should be thinking about what parts are similar and if we can compose any `AsyncElement` components here to reduce the complexity.

src/api/classify_element.rs
```diff
@@ -0,0 +1,349 @@
...
+                    if let Ok(task) = self.wake_providers[port].try_recv() {
+                        task.notify();
+                    }
+                }
+            }
+        }
+    }
+}
+
+/// Split Element Provider, exactly the same as AsyncElementProvider, but
```
ddowl, line 206: `SplitElementProvider`?

CollinValley: Oh needs cleanup, `ClassifyElement` used to be `SplitElement`

ddowl: Can we not modify `AsyncElementProvider` to have the bounds `E: Element + ClassifyElement`? I guess then the bound isn't tight enough to be useful.

src/api/classify_element.rs
```diff
@@ -0,0 +1,349 @@
...
+            },
+            Ok(None) => {
+                Ok(Async::Ready(None))
+            },
+            Err(TryRecvError::Empty) => {
+                let task = task::current();
+                if let Err(_) = self.await_consumer.try_send(task) {
+                    task::current().notify();
+                }
+                Ok(Async::NotReady)
```
ddowl, line 275: This logic is repeated on line 102. We should decompose it into a function.

src/api/classify_element.rs
CollinValley: Maybe,  I just didn't want to muck around with it at the time.  It seems possible. Though Classify doesn't do a transformation,  hence the different trait.  Classify implements `fn classify` which outputs essentially a port number with a packet.

src/api/classify_element.rs
```diff
@@ -0,0 +1,349 @@
...
+    /// ###
+    fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
+        match self.from_consumer.try_recv() {
+            Ok(Some(packet)) => {
+                if let Ok(task) = self.wake_consumer.try_recv() {
+                        task.notify();
+                }
+                Ok(Async::Ready(Some(packet)))
+            },
+            Ok(None) => {
```
ddowl, line 267: It's a shame we can't use `try_ready!` here since we need the custom logic for the `Err(TryRecvError::Empty)` arm.

src/api/classify_element.rs
```diff
@@ -0,0 +1,349 @@
...
+    use crate::api::element::{Element};
+    use crate::utils::test::packet_generators::{immediate_stream, PacketIntervalGenerator};
+    use crate::utils::test::packet_collectors::ExhaustiveCollector;
+    use crossbeam::crossbeam_channel;
+    use core::time;
+
+    use futures::future::lazy;
+
+    #[allow(dead_code)]
+    struct IdentityElement {
```
ddowl, line 296: IdentityElement should definitely be defined elsewhere for testing purposes so we don't have to keep redefining it. We can do that in a separate commit.

src/api/classify_element.rs
ddowl: Maybe @moosingin3space can weigh in?

src/api/classify_element.rs
```diff
@@ -0,0 +1,349 @@
...
+            }
+        }
+    }   
+
+    #[test]
+    fn one_classify_element() {
+        let default_channel_size = 10;
+        let number_branches = 2;
+        let packets = vec![0, 1, 2, 420, 1337, 3, 4, 5, 6, 7, 8, 9, 11];
+        let packet_generator = PacketIntervalGenerator::new(time::Duration::from_millis(100), packets.clone().into_iter());
```
ddowl, line 331: Again for another PR, but I think we should prefer using `immediate_stream` for packet generators to keep tests speedy, and include a test for each component that confirms it can handle pauses in the input stream.

src/api/classify_element.rs
CollinValley: Makes sense.

src/api/classify_element.rs
CollinValley: Yeah I can open an issue

src/api/classify_element.rs
CollinValley: :+1:

ddowl:
Other than a few nitpicks I really like the `ClassifyElement`. It's implementation is _mostly_ straightforward, save for task waking issues (DAMN YOU CROSSBEAM!). The tests show a simple example, but it's descriptive and cool to see it actually work.

That being said, let's get these commits out into different PRs so we can address each of the issues in isolation.

src/api/classify_element.rs
CollinValley: https://github.com/route-rs/route-rs/issues/29 Created issue.

src/api/classify_element.rs
CollinValley: https://github.com/route-rs/route-rs/issues/29 Created issue

src/api/classify_element.rs
CollinValley: https://github.com/route-rs/route-rs/issues/30

src/api/classify_element.rs
CollinValley: https://github.com/route-rs/route-rs/issues/31

src/api/classify_element.rs
CollinValley: https://github.com/route-rs/route-rs/issues/32

ddowl:
Closed since we've reviewed these changes in the meetup today.

PR #27, opened 2019-07-29
graphgen: Generate the pipeline from a graph

scgruber:
This commit includes the binary crate for graphgen, which can be used
to produce pipeline module for use in another crate. It also includes
trivial-identity, which is an example router with code generated by
graphgen.

This commit still needs a lot of work. In particular:
  - [x] There are three in-progress copies of the pipeline example
  - [ ] There are no comments or documentation
  - [x] There are no unit tests
  - [x] The autogen variable names are ugly
  - [x] main.rs should probably be refactored into some more modules


scgruber:
NB: As of right now, the autogen code you should look at is `pipeline_petgraph.rs`. Ignore the other two pipeline files in `trivial-identity`.

scgruber:
Rescoping the following into separate issues for future work:
  - The generator only creates ElementLinks, not AsyncElementLinks
  - Integration test driver
  - There should be some more interesting examples

CollinValley (approved):
This is such a big PR, necessarily, it is hard to sum up.  But overall I really really like this.  It is good, well written code, comes with tests, clear function names, clear division of responsibility.  There are a few places it gets complex, mostly cause rust is verbose when doing string manipulation.  

I'd like to say thanks for all the work you have put in to laying a good foundation for what is likely to be a killer feature if this whole thing gets big some day. 

Approved, you should give a walk-through/demo/Q&A at then next meetup so everyone is up to speed.

PR #28, opened 2019-07-30
Move Elements to seperate modules

CollinValley:
This PR moves the currently defined elements to their own modules.  It doesn't include any of the changes such as the classify and join elements, but we need these changes in to merge the others.


ddowl (approved):
Sorry for the late review. Looks good to me, since I already looked these over in the previous mega-PR

PR #39, opened 2019-08-04
Enforce community style

### LINKED ISSUE #37: Make git commit hook to enforce style standards

ddowl:
We should use Rust's style tools so we don't have to worry about style in CRs. These commits add Githooks to enforce community style standards, an `init.sh` script to help install the tools and configure Githooks for local repos, and applies the style changes.

fixes #37


CollinValley (dismissed):
This all looks really good to me, I would like to get this PR merged ASAP so I can change all the overseer  stuff and such.  Since this touches so many files it will cause merge conflicts, let's get it in now before more people start their work.
@scgruber 
@bweis  Can you two fast-track your reviews on this PR?

.githooks/pre-push
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+branch="$(git rev-parse --abbrev-ref HEAD)"
+
+echo $branch
+
+# Restrict pushing to master
+if [ "$branch" = "master" ]; then
+  echo "You can't commit directly to '"${branch}"' branch"
+  exit 1 # exit with error status to block push
```
scgruber, line 10: This is gonna block me when I try to sync master on my personal fork, I think. We shouldn't be enforcing things like this on the client side.

scgruber (commented):
I'm not personally a fan of client-side hooks for this sort of thing, because I like to write sloppy in-progress commits and then push them later. My preference would be to set up Travis or whatever to run these checks against PRs and let people run `cargo fmt` and `cargo clippy` manually on their local end if desired.

.githooks/pre-push
CollinValley: Yeah I guess that part isn't really needed, since $master branch and origin/master are not really the same thing

CollinValley: Per the commit hooks, I actually do like them. Just like running shellcheck before each commit, it will be annoying workflow to constantly have to have the PRs get first pass rejected when people forget the cargo fmt and cargo clippy. It is also official language layout guidelines, so it isn't like these choices are just out of the blue.

CollinValley: I view the cargo fmt and clippy like running cargo build and cargo test before commiting code. I don't commit code that doesn't compile.  And if you really want to you can disable the hook for the commit.

scgruber: Maybe put the checks in pre-push then? I don't want them when I'm making commits locally just trying to keep track of my work. But maybe other people feel strongly about this. I likely won't init the hooks in any case, so probably just ignore me.

CollinValley (commented):
Let's shelve this until the meetup, good light shed-based conversation can be had about this.  Also the current PR needs rework since the codebase is now linted in master.

.githooks/pre-push
ddowl: I know in codebases that have pre-commit lint checks, I use `--no-verify` to skip them if I want to make an intermediate "checkpoint" commit that I'll never push. @scgruber maybe `fmt` and `clippy` would be better pre-push, and definitely should be included in Travis. I'd like to have them on the client side somehow to avoid situations where the initial pass of reviews consists of '"dude, you didn't `fmt` your code", "my b"'. 

I'm ok with removing the push-to-master restriction since not everyone uses a feature branch workflow so strictly. As an alternative, we could restrict the script to gate remote pushes, which is the actual intention.

CollinValley: I think the compromise that we move all the checks to be on push, both checking fmt and clippy, as well as preventing pushes directly to master, would be a good solution.

ddowl:
Resolved by #76

Issue #24, opened 2019-07-24
AsyncElement: Test more edge cases


CollinValley:
We need move tests to figure out if there are any additional bugs with our AsyncElement.  We need help writing more tests that do more interesting things.  Even basic cases like a stream that immediately returns None as input. Or seeing if there are any tests that should panic and writing those.

PR #40, opened 2019-08-04
AsyncElement: add trivial tests

eeverett6:
Add tests to AsyncElement for two trivial cases, an empty stream that
immediately returns Ready(None), and an AsyncElementLink created with an
empty channel that should panic.

This commit makes progress on #24, but does not complete it.

### DIFF
--- a/src/api/async_element.rs
+++ b/src/api/async_element.rs
@@ -24,6 +24,9 @@ pub struct AsyncElementLink<E: AsyncElement> {
 
 impl<E: AsyncElement> AsyncElementLink<E> {
     pub fn new(input_stream: ElementStream<E::Input>, element: E, queue_capacity: usize) -> Self {
+        if queue_capacity == 0 {
+            panic!("queue capacity must be non-zero")
+        }
         let (to_provider, from_consumer) =
             crossbeam_channel::bounded::<Option<E::Output>>(queue_capacity);
         let task_park: Arc<AtomicCell<TaskParkState>> =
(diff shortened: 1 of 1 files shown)


src/api/async_element.rs
```diff
@@ -352,6 +352,44 @@ mod tests {
...
+            Ok(())
+        }));
+
+        let router_output: Vec<_> = r.iter().collect();
+        assert_eq!(router_output, packets);
+    }
+
+    #[test]
+    #[should_panic]
+    fn one_async_element_with_empty_channel() {
```
ddowl: Not sure if this test is complete yet. What is the empty channel here we're testing?

src/api/async_element.rs
```diff
@@ -352,6 +352,44 @@ mod tests {
         assert_eq!(router_output.len(), 2001);
     }
 
+    #[test]
+    fn one_async_element_with_empty_stream() {
```
ddowl: This is the right idea. We have an issue open to abstract away some of the tokio running and assertion code away, but you'll necessarily have to define the input stream for your router, configure it via Element links, and assert that some stuff happened to the output stream.

src/api/async_element.rs
CollinValley: we are testing that if the `default_channel_size` is set to 0, it panics.

src/api/async_element.rs
CollinValley: This looks good to me, does what we expect.

CollinValley (dismissed):
I think this can be merged as is. Everything looks good to me. Just resubmit the PR and we are off to the races.

CollinValley:
Or if you want to layer on more changes to fully resolve #24 then just add them on top, your first commit looks good.

ddowl (approved):

PR #42, opened 2019-08-07
Fix deadlock issue for Async Element

CollinValley:
This is a draft because I am excited, but I managed to resolve the deadlock issue and remove the overseer though the use of AtomicCell instead of Channels to pass tasks around, had to also define an enum to deal with teardown behavior.  Check it out,  fridge is just a reference to the milk in fridge async locking problem.  That name won't make the final draft.

### DIFF
--- a/src/api/async_element.rs
+++ b/src/api/async_element.rs
@@ -1,7 +1,14 @@
 use futures::{Future, Stream, Async, Poll, task};
 use crossbeam::crossbeam_channel::{Sender, Receiver, TryRecvError};
 use crossbeam::crossbeam_channel;
+use crossbeam::atomic::AtomicCell;
 use crate::api::ElementStream;
+use std::sync::Arc;
+
+enum FridgeStatus {
+    Dead,
+    Alive(Option<task::Task>)
+}
 
 pub trait AsyncElement {
     type Input: Sized;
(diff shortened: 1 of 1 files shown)


src/api/async_element.rs
```diff
@@ -302,7 +275,7 @@ mod tests {
     }
 
     #[test]
-    fn one_async_element() {
+    fn one_async_element<'a>() {
```
ddowl, line 278: Necessary lifetime?

ddowl:
This is sick. Solid milk management 👌 👌 👌 👌. For those not at the last meetup, we talked about how managing the task waking is similar to the [too much milk problem](https://people.cs.umass.edu/~emery/classes/cmpsci377/f07/scribe/scribe8-1.pdf). It seems like the `AtomicCell` is the right call here since you can atomically [swap](https://docs.rs/crossbeam/0.5.0/crossbeam/atomic/struct.AtomicCell.html#method.swap) values into and out of it.

src/api/async_element.rs
CollinValley: Oh yeah, wasn't needed ,solved with Arc, will catch that on cleanup, also when I change the names and such.

CollinValley:
Taking suggestions for what to call task_fridge so it isn't confusing.  task_rendezvous, task_bed, task_wake? 
Also the enum could be refactored to be cases {Dead, Empty, Full(task)} instead of {Dead, Alive(`Option<Task>`)}.

CollinValley:
Closing this, official PR for this issue can be found at #46

PR #43, opened 2019-08-07
api: pub use to reduce namespace clutter

scgruber:
It's nice for us to group parts of the API into different files, but
this means that the namespaces are getting overly verbose. We can
declare our subfiles with pub use to export them into the main api
namespace and make it easier to work with.

### DIFF
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -1,8 +1,15 @@
 use futures::Stream;
 
-pub mod element;
-pub mod async_element;
-pub mod classify_element;
-pub mod join_element;
+mod element;
+pub use self::element::*;
+
+mod async_element;
+pub use self::async_element::*;
+
+mod classify_element;
+pub use self::classify_element::*;
+
+mod join_element;
+pub use self::join_element::*;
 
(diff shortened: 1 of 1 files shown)


CollinValley:
If it doesn't cause collisions I don't see why not.  Especially since we already call the links, consumers, and providers by unique names. Though we may not want to expose every function down the road, might be better to explicitly name the things we want exported rather than *?  I wish I was better versed in how rust does these things.

CollinValley:
I guess my question is, if we declare a non-public struct, or function in the async_elemnt.rs file. Will this then export that as public, or does it only export the pub functions to the main api?

ddowl:
👍. It's useful to have modules internally to separate out the different classes of elements, but to a client it doesn't necessarily matter, and referencing `route-rs::api::ElementLink` or even `route-rs::ElementLink` would be preferable. One potential change could be an `route-rs::elements` top-level module if we anticipate non-element modules appearing in the API.

CollinValley (approved):
I approve this as is.  Right now we don't have any private functions, so if that is an issue in the future, we can change it in the future.  I want to keep PRs small and steady anyways. I think this is a good change.

CollinValley:
@scgruber  I will let you do the honors of merging it.

PR #44, opened 2019-08-08
Add Travis CI configuration

scgruber:
This is a pretty basic configuration. It defines three stages:
  - lint - runs rustfmt and clippy to tell us if we are writing
      good code.
  - build - actually builds the crate
  - test - runs the tests

Right now this is a single-crate build. We'll have to change some
of the commands when we refactor into a workspace project.


CollinValley:
I'm a bit confused if this is correct because the Travis CI failed it's own PR?  In concept is seems good to me. Gives us a good basic starting platform, we can refactor over time to fit our needs.

scgruber:
Note: this PR duplicates the style fixups from #39

CollinValley (commented):
Looks good to me, going to merge this.  Side effect is people's current stacks are going to break and need rebase, but that is fine.  Better to bite the bullet early.  #39 will need some further discussion then whether we want to enforce that on commit by default.

CollinValley (approved):
Forgot to approve, see above comment.

PR #46, opened 2019-08-13
TaskPark: Remove need for overseer on Async and Classify Elements

CollinValley:
This PR changes how we pass task handles between the consumer and provider in our Asynchronous element links from channels to AtomicCells.  In previous implementations, we had to work around a deadlock in the task channel by adding an overseer to handle waking of stuck tasks. This is no longer necessary since the deadlock state where both the consumer and provider are asleep awaiting each other can not happen with them sharing one location to park their task handles.

Note that TaskParks can be in 3 possible states.
1: Full, the task park contains a task handle
2: Empty, the task park does not contain a task handle, but you can park one here.
3: Dead, one of the sides is no longer active, and as such it is no longer safe to park a task handle here.

Once I have a functioning join implementation with task parks, will make a second pass at cleaning up the code. I think there are a few easy places to simplify logic and add helper functions, but I don't want to invest the time just yet when the task park may require extensions to work in broader contexts.

### DIFF
--- a/src/api/async_element.rs
+++ b/src/api/async_element.rs
@@ -1,7 +1,10 @@
 use crate::api::ElementStream;
+use crate::api::TaskParkState;
+use crossbeam::atomic::AtomicCell;
 use crossbeam::crossbeam_channel;
 use crossbeam::crossbeam_channel::{Receiver, Sender, TryRecvError};
 use futures::{task, Async, Future, Poll, Stream};
+use std::sync::Arc;
 
 pub trait AsyncElement {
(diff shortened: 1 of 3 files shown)


scgruber (approved):
So much delete

ddowl:
***_DOIN A BIG OL DELETE_***

It's not entirely obvious why we have to include a `TaskPark::Dead` state. Why is that?

ddowl:
Whoops, gotta merge it first.

PR #47, opened 2019-08-13
api: make tests use immediate_stream

### LINKED ISSUE #31: Prefer ImmediateStream for tests.
Unless a different stream is needed to accurately test, we should prefer ImmediateStream in our basic test cases since it runs quicker, allowing us to run the the suite of tests in less time.

ddowl:
Most tests should use immediate_stream as the default packet_generator
to ensure that they are _speedy_.

This change doesn't prune`LinearIntervalGenerator` or `ExhaustiveDrain` since they may still be useful.

fixes #31

### DIFF
--- a/src/api/async_element.rs
+++ b/src/api/async_element.rs
@@ -281,10 +281,7 @@ mod tests {
     fn one_async_element() {
         let default_channel_size = 10;
         let packets = vec![0, 1, 2, 420, 1337, 3, 4, 5, 6, 7, 8, 9];
-        let packet_generator = PacketIntervalGenerator::new(
-            time::Duration::from_millis(100),
-            packets.clone().into_iter(),
-        );
+        let packet_generator = immediate_stream(packets.clone());
 
         let elem0 = AsyncIdentityElement { id: 0 };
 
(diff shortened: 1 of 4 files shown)


scgruber (commented):
I don't understand why we're keeping tests that have waits in them? This goes against the general testing principle of not sleeping or waiting, so I'd rather we didn't.

ddowl:
I'm including them so we can be sure that each type of element can deal with bursty streams of packets, and handles the `Some(Async::NotReady)` case. It's the easiest way to invoke `Some(Async::NotReady)` from the packet flow, since a direct stream (where each call to `poll` yields a packet), always returns `Some(Async::Ready)` until the stream ends, and returns `None`. Introducing a bursty packet flow can also reveal weird synchronization issues between `Future`s.

Its true that it makes the test slower, but unfortunately there's not a great way to ensure that `poll` is properly tested otherwise. We can always reduce the wait interval, but then we wouldn't want to lower it so much that the `NotReady` case never gets invoked.

ddowl:
If we can think of a faster way to test the `NotReady` cases, we can reserve `PacketIntervalGenerator` and variants for slower integration tests.

CollinValley (commented):
I approve of this, let's reduce the interval to 10ms instead of 100, accomplishes the same thing, but it means the tests will complete a lot faster. Cargo test is multi-threaded so if we keep everything resolving in under a second that would be quick enough.

ddowl:
Updated wait times.

CollinValley (dismissed):
Looks good

CollinValley:
Merged this PR against master, cleanup up some small build errors.

CollinValley (approved):

PR #49, opened 2019-08-13
Join Element: Switch to TaskPark instead of Overseer

### LINKED ISSUE #48: Join Element: Task waking without Overseer
Much like the task_park used in the classify and async elements, we want to be able to use the task_park for thread sleeping in the join element.

CollinValley:
This PR changes Join Element to use the TaskPark method instead of the Overseer method of preventing deadlocks. In general, it works the same as Async or Classify TaskParks. The only primary difference is that when the provider goes to sleep, it hands it task handle to ever consumer. This may be wasteful since it means the provider is now being over-scheduled. This is still far better than the Overseer though.  See commit message for more. This resolves #48

### DIFF
--- a/src/api/join_element.rs
+++ b/src/api/join_element.rs
@@ -1,12 +1,14 @@
 use crate::api::ElementStream;
+use crate::api::TaskParkState;
+use crossbeam::atomic::AtomicCell;
 use crossbeam::crossbeam_channel;
-use crossbeam::crossbeam_channel::{Receiver, Sender, TryRecvError};
+use crossbeam::crossbeam_channel::{Receiver, Sender};
 use futures::{task, Async, Future, Poll, Stream};
+use std::sync::Arc;
 
 pub struct JoinElementLink<Packet: Sized> {
     pub consumers: Vec<JoinElementConsumer<Packet>>,
     pub provider: JoinElementProvider<Packet>,
-    pub overseer: JoinElementOverseer<Packet>,
 }
 
 impl<Packet: Sized> JoinElementLink<Packet> {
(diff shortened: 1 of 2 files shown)


CollinValley:
Rebased this branch on master after Ethan's changes went in.

src/api/join_element.rs
```diff
@@ -302,8 +248,28 @@ impl<Packet: Sized> Stream for JoinElementProvider<Packet> {
             }
         }
 
-        let task = task::current();
-        if self.await_consumers.try_send(task).is_err() {
+        // For now, we are going to store our task handle in every task_park, since we do not have a good way to let
+        // the cosumers share one big task park. Perhaps we could use a mutexed queue, or something similar, rather than
+        // spraying the task handle to every consumer. I think the current implementation results in the provider being
+        // scheduled quite frequently.
+        let provider = task::current();
```
ddowl, line 255: `provider_task`?

ddowl (dismissed):
I'd squash those commits together and check the commit message again before merging, but this all lgtm.

Really happy with how `TaskParkState` ended up.

src/api/join_element.rs
CollinValley: I left it as provider and consumer rather than provider_task and consumer_task, cause provider.notify reads better.

CollinValley:
Fixed some typos and clarified language in the commit message. Kept the Async Element changes in their own commit so that we are not tying the modifications together when they don't have to be.

ddowl:
Oh whoops, I didn't realize those were different files in each commit. LGTMF

ddowl (approved):
LGTMMD <sub><sup>Looks Good To Me My Dude</sup></sub>

PR #50, opened 2019-08-14
ClassifyElement: simplify classify signature

### LINKED ISSUE #35: Improve Classify Function
The Classify function in ClassifyElement could probably just return the port the packet should be output on, rather than a tuple of (port, packet).

ddowl:
Modifies the signature of the `classify` function to take a borrow of a
packet (since it should only need to be observed), and return only the
port it should be dispatched to.

fixes #35


CollinValley (approved):
LGTMMD 
Pretty simple change.

Issue #30, opened 2019-07-30
Move test element definitions to common location.


CollinValley:
Some testing elements, like our IdentityElement, keep being redefined in each element link module.  Let's define it in a common location.

PR #51, opened 2019-08-14
IdentityElement: move duplicated element to test

ErrHu:
This deduplicates IdentityElement and moves AsyncIdentityElement
with it to identity_elemts.rs. They were also made more generic
so it's not limited to i32. We may want to move these out of the
test directory if they become used outside tests.

Also removed dead_code markers

Issue #30


CollinValley (approved):
I like this change, I'm going to merge it as is.  I'll create two new issues to switch Classify and Join over to the new element test framework.

Issue #29, opened 2019-07-30
Investigate combining common logic in Asynchronous links


CollinValley:
We noticed that among AsyncElementLink, ClassifyElementLink, and JoinElementLink,  their subcomponents such as their consumers and producers, contain a lot of reoccurring logic.  We should try to compose this in a helper function.

ddowl:
Leaving this issue open since there might be more possible refactoring among concurrent elements. This work might just fall under #32 though, in which case we can close this issue.

PR #54, opened 2019-08-14
task_park: extract common utilities

ddowl:
It turns out that there are very repeatable patterns to dealing with
task-wake management in concurrent elements. Those are:

1) unpark_and_notify: "I have work to, so don't bother waking me up."
2) park_and_notify: "I don't have any work to do. Let me know when I do."
3) die_and_notify: "I'm shutting down, and I'll either let you wake me
up if you're around, or I'll self-notify if I die".

park and unpark variants will notify a task if it exists in the park as
the new state is swapped in, and will maintain the "dead" state so
die_and_notify works appropriately.

relates to #29

### DIFF
--- a/src/api/async_element.rs
+++ b/src/api/async_element.rs
@@ -1,9 +1,9 @@
+use crate::api::task_park::*;
 use crate::api::ElementStream;
-use crate::api::TaskParkState;
 use crossbeam::atomic::AtomicCell;
 use crossbeam::crossbeam_channel;
 use crossbeam::crossbeam_channel::{Receiver, Sender, TryRecvError};
-use futures::{task, Async, Future, Poll, Stream};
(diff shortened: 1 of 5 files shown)


ddowl:
Forgot to run `cargo fmt`

CollinValley:
> Forgot to run `cargo fmt`

Lol need that pre-push hook

ddowl:
> > Forgot to run `cargo fmt`
> 
> Lol need that pre-push hook

:sob:

CollinValley (approved):
This is exactly what I was thinking we should do when I was writing this code.  I agree the small weirdness in the join element is worthwhile to reduce complexity. Especially since we only run into that case when the link is in tear down. All in all this is :100: :+1:  :hankey:

PR #55, opened 2019-08-14
Async Element: Shorten longest test with immediate stream

CollinValley:
This is a small change to one of our tests that is rather long
running, at least compared to other tests. Switching to immediate
stream when we are not testing packet delay improves the speed.

Test run time reduces from 1.349s -> 0.061s

All tests now run on my build machine in ~0.3s, check my flex.


ddowl (approved):
_GOTTEM_

PR #57, opened 2019-08-14
ClassifyElement: more tests

### LINKED ISSUE #25: ClassifyElement: Test more edge cases
We should add more tests, specifically we need tests that contain more output branches, have longer to compute split functions, have multiple layers of channels. etc etc.

ddowl:
I noticed that certain router configs would either panic or not run at
all when not referencing the last `provider` in a `ClassifyElementLink`
via `pop`. That might require more investigation as we write our POC
integration tests.

resolves #25

### DIFF
--- a/src/api/classify_element.rs
+++ b/src/api/classify_element.rs
@@ -239,8 +239,29 @@ mod tests {
         }
     }
 
+    #[allow(dead_code)]
+    struct ClassifyFizzBuzzElement {
+        id: i32,
+    }
+
+    impl ClassifyElement for ClassifyFizzBuzzElement {
+        type Packet = i32;
+
+        fn classify(&mut self, packet: &Self::Packet) -> usize {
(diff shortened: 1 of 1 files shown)


src/api/classify_element.rs
```diff
@@ -318,46 +337,149 @@ mod tests {
...
+    }
+
+    #[test]
+    fn fizz_buzz_to_even_odd() {
+        let default_channel_size = 10;
+        let packet_generator = immediate_stream(0..=30);
+
+        let fizz_buzz_elem = ClassifyFizzBuzzElement { id: 0 };
+
+        let mut fizz_buzz_elem_link: ClassifyElementLink<ClassifyFizzBuzzElement> =
```
eeverett6, line 437: Is the fully specified type needed here because we have two different ClassifyElementLinks?

bweis (approved):
LGTM

PR #58, opened 2019-08-14
fix: decrease packet interval from 100 to 10

ddowl:
This was missed during a rebase or something.


ddowl:
Closing since @CollinValley 's got this in #55.

PR #66, opened 2019-08-15
Pipeline and example

scgruber:
This PR implements the pipeline structure for running a router in a binary crate, and adds an example. It also moves the IdentityElement and AsyncIdentity element to runtime::element so they can be used beyond runtime's tests.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,4 +1,10 @@
 [workspace]
 members = [
+    # API Crates
+    # These crates can actually be used by other people and might be published to crates.io
     "route-rs-runtime",
+
+    # Example crates
+    # These crates show usage of route-rs features and should _not_ be published to crates.io
+    "examples/trivial-identity",
 ]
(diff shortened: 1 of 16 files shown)


CollinValley:
First pass on this: It looks really cool, I'm going to take deeper look tonight/tomorrow. At first glance I am excited about the pipeline concept. It can really wrap up a lot of complexity.

CollinValley (dismissed):
Looking at this, it all looks good. We should merge it. There will be more work down the road to make it more useful, but this is a good abstraction to contain a bunch of links and elements. Perhaps common sets of elements can just be abstracted into pipelines in the future. We may consider making the boundaries of pipelines futures then, so that they can be chained together, much like a link is. But for right now, I get the purpose of pipeline. 

We don't have to do it in this PR, but I think we should consider moving input and output links into the links folder, rather than leaving them in the pipeline folder. It makes sense that links should stay together, and input and output are just generic channel-to-future, and future-to-channel converters.

CollinValley:
Looks like you are failing clippy

scgruber:
@CollinValley can you review again, now passing tests.

CollinValley (approved):

PR #67, opened 2019-08-17
Join Element: Prevent overscheduling task

CollinValley:
In previous implementations of Task Park, we had an issue where we could
not hand out a common task handle to every consumer when the provider
was going to sleep. In this case, the provider with get scheduled pretty
often, which was pretty inefficient. In this solution I changed the top
level TaskParkState enum to allow it to store either a direct or indirect reference to
the task handle. This enables sharing a common task handle among multiple tasks,
preventing the over-scheduling issue. Additionally, I refactored the task_park helper
functions to minimize redundancy.

### DIFF
--- a/route-rs-runtime/src/link/join_link.rs
+++ b/route-rs-runtime/src/link/join_link.rs
@@ -3,7 +3,7 @@ use crate::link::PacketStream;
 use crossbeam::atomic::AtomicCell;
 use crossbeam::crossbeam_channel;
 use crossbeam::crossbeam_channel::{Receiver, Sender};
-use futures::{Async, Future, Poll, Stream};
+use futures::{task, Async, Future, Poll, Stream};
 use std::sync::Arc;
 
 pub struct JoinElementLink<Packet: Sized> {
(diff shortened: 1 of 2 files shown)


scgruber:
Have we not just switched the overnotify problem from the provider to the consumer? Now if multiple consumers go to sleep, they will keep waking each other up because all of their notifies are going into the same cell.

CollinValley:
> Have we not just switched the overnotify problem from the provider to the consumer? Now if multiple consumers go to sleep, they will keep waking each other up because all of their notifies are going into the same cell.

It would be an issue if all the consumers shared a single task park. They all have their own, and they directly park in it. So when the provider wishes to sleep it posts its task indirectly in each of the consumers task parks, it iterates over each task park per consumer, awakens any that are asleep, and clones the reference.

 Keep in mind the consumers only park their task when they encounter a full queue. And providers only park their task when encountering entirely empty queues.  This whole setup is rather confusing because there are a lot of states that are covered but are extremely unlikely, or even impossible, to happen.  When the provider is parking its task, it is unlikely to encounter a full task park, since each time it pulls a packet from the queue it will notify any task handle there. This means the process of emptying the queue will notify the consumer, they may not be scheduled yet, but they are in the task queue. 

The key to keeping everything efficient is only one task handle is created per NotReady returned.  The provider can hand out multiple references to the same task handle object using the Arc<AtomicCell<_>>, but it comes at a cost of having to do a double dereference. Consumers, since they only need to park in one location, just use the 'direct' park, called 'park_and_notify()'

scgruber:
Oh I see, makes sense

CollinValley:
Yeah it gets pretty gnarly,  needed a whiteboard to figure out how to combine all the logic.

scgruber (approved):

PR #76, opened 2019-08-21
add pre-push style hooks

ddowl:
Client side style hooks can be initialized by running `init-dev.sh`. It
essentially runs the same style checks that Travis will run before
accepting a PR, but runs on the client so you can avoid the shame of
force-pushing style fixes before a review.


CollinValley (approved):
LGTMMD

PR #78, opened 2019-08-24
Change names: Ingressor, Egressor, [type]Link

CollinValley:
Ben Weis is having some trouble getting this done on his macbook due to the tests not running for some reason. Something due to not enough file descriptors to run 20+ instances of tokio at once.  So I just went ahead with his naming suggestion and implemented the concept. Added him as a reviewer.  I also removed unnecessary Element words from things.

### DIFF
--- a/route-rs-graphgen/src/main.rs
+++ b/route-rs-graphgen/src/main.rs
@@ -15,7 +15,7 @@ use std::collections::HashMap;
 mod codegen;
 mod pipeline_graph;
 
-enum ElementLink {
+enum Link {
     Input,
     Output(pipeline_graph::XmlNodeId),
     Sync(pipeline_graph::XmlNodeId, pipeline_graph::XmlNodeId),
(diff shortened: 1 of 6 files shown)


route-rs-graphgen/src/main.rs
```diff
@@ -15,7 +15,7 @@ use std::collections::HashMap;
 mod codegen;
 mod pipeline_graph;
 
-enum ElementLink {
+enum SyncLink {
```
scgruber: This actually should be generic, I just don't have other cases yet.

CollinValley: Okay, should the name be changed back? Or should it just be Link?

scgruber: It could just be `Link`, if that doesn't feel too terse.

ddowl (approved):
Wow, these PRs are getting big. Happy with these new names, they're much more clear

PR #79, opened 2019-08-25
Rewrite ClassifyElementLink to work better in pipelines

### LINKED ISSUE #77: Write an example crate that uses Join and Classify

scgruber:
This commit rewrites ClassifyElementLink to work more similarly to the
standard ElementLink. This allows us to define the dispatch from it
inline in the pipeline, which makes code generation easier in graphgen.

This commit also restructures ClassifyElement as a subtrait of Element,
which keeps the structure more consistent for our users.

examples/dns-interceptor shows the usage of ClassifyElementLink and
JoinElementLink, so we know that they mesh with the overall pipeline
implementation. A future commit will support this example with graphgen.

Fixes #77

### DIFF
--- a/examples/dns-interceptor/src/elements.rs
+++ b/examples/dns-interceptor/src/elements.rs
@@ -1,0 +1,99 @@
+use crate::packets::*;
+use route_rs_runtime::element::{ClassifyElement, Element};
+use std::collections::HashMap;
+
+pub struct SetInterfaceByDestination {
+    lan_subnet_prefix: u32,
+    lan_subnet_mask: u32,
+}
+
+impl SetInterfaceByDestination {
+    pub fn new(lan_subnet_prefix: u32, lan_subnet_mask: u32) -> Self {
+        SetInterfaceByDestination {
+            lan_subnet_prefix,
+            lan_subnet_mask,
+        }
+    }
+}
+
+impl Element for SetInterfaceByDestination {
+    type Input = (Interface, SimplePacket);
+    type Output = (Interface, SimplePacket);
+
+    fn process(&mut self, packet: Self::Input) -> Self::Output {
+        let dest_ip = packet.1.destination.ip;
+        if (dest_ip & self.lan_subnet_mask) == self.lan_subnet_prefix {
+            (Interface::LAN, packet.1)
+        } else {
+            (Interface::WAN, packet.1)
+        }
+    }
+}
+
+#[derive(Debug, PartialEq)]
+pub enum ClassifyDNSOutput {
+    DNS,
+    Other,
+}
+
+pub struct ClassifyDNS {}
+
+impl ClassifyDNS {
+    pub fn new() -> Self {
+        ClassifyDNS {}
+    }
+}
+
+impl Element for ClassifyDNS {
+    type Input = (Interface, SimplePacket);
+    type Output = (ClassifyDNSOutput, Self::Input);
+
+    fn process(&mut self, packet: Self::Input) -> Self::Output {
+        match packet.1.destination.port {
+            53 => (ClassifyDNSOutput::DNS, packet),
+            _ => (ClassifyDNSOutput::Other, packet),
+        }
+    }
+}
+
+impl ClassifyElement for ClassifyDNS {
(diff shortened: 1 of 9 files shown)


route-rs-runtime/src/link/classify_link.rs
```diff
@@ -7,15 +7,16 @@ use crossbeam::crossbeam_channel::{Receiver, Sender, TryRecvError};
...
     pub providers: Vec<ClassifyElementProvider<E>>,
 }
 
-impl<E: ClassifyElement> ClassifyElementLink<E> {
+impl<'a, E: ClassifyElement> ClassifyElementLink<'a, E> {
     pub fn new(
-        input_stream: PacketStream<E::Packet>,
+        input_stream: PacketStream<E::Input>,
         element: E,
+        dispatcher: Box<dyn Fn(E::Class) -> usize + Send + Sync + 'a>,
```
CollinValley, line 19: I wish this did not have to be dynamic, but since the eliminates an element to set the type, which also would include a dyn function call (The poll), this is a reasonable cost to pay, maybe in the future we can remove this? Idk might just be a small cost of being generic.

CollinValley (commented):
I approve these changes, but I want to merge my name change first so my I don't get so many merge conflicts. It will be easier for you to fix the names in your code that for me.

examples/dns-interceptor/src/elements.rs
```diff
@@ -0,0 +1,100 @@
...
+        LocalDNSInterceptor { intercept_rules }
+    }
+}
+
+impl Element for LocalDNSInterceptor {
+    type Input = (Interface, SimplePacket);
+    type Output = (Interface, SimplePacket);
+
+    fn process(&mut self, packet: Self::Input) -> Self::Output {
+        println!("Processing {:?}", packet);
```
ddowl: Intentional?

route-rs-runtime/src/link/classify_link.rs
```diff
@@ -111,16 +121,17 @@ impl<E: ClassifyElement> Future for ClassifyElementConsumer<E> {
...
                 }
             }
-            let packet_option: Option<E::Packet> = try_ready!(self.input_stream.poll());
+            let packet_option: Option<E::Input> = try_ready!(self.input_stream.poll());
 
             match packet_option {
                 None => return Ok(Async::Ready(())),
                 Some(packet) => {
-                    let port = self.element.classify(&packet);
+                    let (class, actual_output) = self.element.process(packet);
```
ddowl, line 127: Why does the new `ClassifyElement`'s `process` need to return `output`? Can not not force it to borrow `packet`? The interface becomes a little clumsier now, since we aren't allowing any change from `packet` to `actual_output`.

CollinValley:
After some discussion with Sam, we can merge these changes after the conflicts are resolved. Drew has that comment about the Print being there that probably shouldn't be?  But while we don't like the closure, it is fine for now until we replace it with a trait-defined function called dispatch that does essentially the same thing. We can do that later though, the PoC router is more important.

CollinValley (dismissed):
Approved, need to fix something so Travis is okay, I think you need to add a license to your example crate.

route-rs-runtime/src/element/base_element.rs
```diff
@@ -12,8 +12,14 @@ pub trait AsyncElement {
...
 
-pub trait ClassifyElement {
-    type Packet: Sized;
-
-    fn classify(&mut self, packet: &Self::Packet) -> usize;
+pub trait ClassifyElement:
+    Element<
+    Output = (
+        <Self as ClassifyElement>::Class,
+        <Self as ClassifyElement>::ActualOutput,
```
ddowl, line 19: We discussed this on Discord, but classification shouldn't perform transformation. We should prefer element composition rather than complicating a base element's interface.

CollinValley: Yeah Sam and I are shelving this issue for this sprint, to unblock him while we work on the POC. Soon after this is merged we will switch the trait composition to rectify this. But the new interface will take some thought so we want to give it some time when we are more free.

CollinValley (approved):

PR #82, opened 2019-08-26
Add checks for dependency license compatibility

scgruber:
Since route-rs is distributed under the MIT license, we want to make
sure that all of our dependencies are similarly nonrestrictive.
Our users will have to treat our package like the most restrictive
license in our dependency tree when deciding if it is usable in their
projects, so we should make this an easy decision for them.

This commit uses cargo-deny to enforce licensing, and updates our own
crates to properly record license information. We also have to
whitelist LICENSE-THIRD-PARTY in crossbeam-channel because it confuses
the scanner. I have manually inspected that license and it is just a
concatenation of licenses that are in our whitelist.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,9 @@ jobs:
     - stage: lint
       install: rustup component add clippy
       script: cargo clippy --all-targets -- -D warnings
+    - stage: lint
+      install: cargo install -f cargo-deny
+      script: cargo deny check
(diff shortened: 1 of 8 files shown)


CollinValley (approved):
Good idea here, don't want to accidentally restrict our work. We chose MIT to specifically be friendly to business if they want to use it.

PR #84, opened 2019-08-26
TransformElements: new elem w/ different I/O types

### LINKED ISSUE #63: Create test elements with different input and output types
For testing, we should have an element that takes in an `int` and outputs a `char`, for example.

eeverett6:
Add a new trivial element for testing that has a different input and
output type. Use std::convert::From to do so generically.

Closes #63

### DIFF
--- a/route-rs-runtime/src/element/transform_elements.rs
+++ b/route-rs-runtime/src/element/transform_elements.rs
@@ -1,0 +1,40 @@
+use crate::element::{AsyncElement, Element};
+use std::convert::From;
+use std::marker::PhantomData;
+
+/*
+  TransformElement
+  This is an element that takes type A and passes the equivalent type B using From
+*/
+#[derive(Default)]
+pub struct TransformElement<A: Sized, B: Sized> {
+    phantom_in: PhantomData<A>,
+    phantom_out: PhantomData<B>,
+}
+
+impl<A, B> TransformElement<A, B> {
+    pub fn new() -> TransformElement<A, B> {
+        TransformElement {
+            phantom_in: PhantomData,
+            phantom_out: PhantomData,
+        }
+    }
+}
+
+impl<A, B: From<A>> Element for TransformElement<A, B> {
+    type Input = A;
+    type Output = B;
+
+    fn process(&mut self, packet: Self::Input) -> Self::Output {
+        Self::Output::from(packet)
+    }
+}
+
+impl<A, B: From<A>> AsyncElement for TransformElement<A, B> {
+    type Input = A;
+    type Output = B;
+
+    fn process(&mut self, packet: Self::Input) -> Self::Output {
+        Self::Output::from(packet)
+    }
+}
--- a/route-rs-runtime/src/link/async_link.rs
+++ b/route-rs-runtime/src/link/async_link.rs
@@ -209,7 +209,7 @@ impl<E: AsyncElement> Stream for AsyncEgressor<E> {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::element::{AsyncIdentityElement, IdentityElement};
+    use crate::element::{AsyncIdentityElement, IdentityElement, TransformElement};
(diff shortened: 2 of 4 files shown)


route-rs-runtime/src/element/transform_elements.rs
```diff
@@ -0,0 +1,45 @@
+use crate::element::{AsyncElement, Element};
+use std::convert::From;
+use std::marker::PhantomData;
+
+/* TransformElement
```
ddowl: nit: Can we keep comment text within newlines of the boundaries like

```
/*
  Yo here's a really important comment
*/
```

route-rs-runtime/src/element/transform_elements.rs
```diff
@@ -0,0 +1,45 @@
...
+        TransformElement {
+            phantom: PhantomData,
+            phantom_out: PhantomData,
+        }
+    }
+}
+
+impl<A, B> Element for TransformElement<A, B>
+where
+    B: From<A>,
```
ddowl: It might be cleaner to inline this trait bound like `TransformElement<A, B: From<A>>`

route-rs-runtime/src/element/transform_elements.rs
```diff
@@ -0,0 +1,45 @@
...
+    type Output = B;
+
+    fn process(&mut self, packet: Self::Input) -> Self::Output {
+        Self::Output::from(packet)
+    }
+}
+
+impl<A, B> AsyncElement for TransformElement<A, B>
+where
+    B: From<A>,
```
ddowl: Ditto L25

route-rs-runtime/src/link/async_link.rs
```diff
@@ -419,4 +419,30 @@ mod tests {
         let router_output: Vec<_> = r.iter().collect();
         assert_eq!(router_output, packets);
     }
+
+    #[test]
+    fn one_async_transform_element() {
+        let default_channel_size = 10;
+        let packets = vec!['r', 'o', 'u', 't', 'e', '-', 'r', 's'];
```
ddowl: `"route-rs".chars().collect::<Vec<char>>()`?

ddowl: Or just `"route-rs".chars()` since `immediate_stream` needs an `Iter`.

route-rs-runtime/src/element/transform_elements.rs
```diff
@@ -0,0 +1,45 @@
+use crate::element::{AsyncElement, Element};
+use std::convert::From;
+use std::marker::PhantomData;
+
+/* TransformElement
+  This is an element that takes type A and passes the equivalent type B using From
+*/
+#[derive(Default)]
+pub struct TransformElement<A: Sized, B: Sized> {
+    phantom: PhantomData<A>,
```
ddowl: I suppose the `PhantomData`s are necessary because of [unused type params](https://doc.rust-lang.org/std/marker/struct.PhantomData.html#unused-type-parameters)?

route-rs-runtime/src/element/transform_elements.rs
eeverett6: Done, though we may want to update this in other places too, I copied this block from identity_elements.rs

route-rs-runtime/src/element/transform_elements.rs
eeverett6: cargo check only worked with the bound after the impl keyword, is that the correct place or does it need to be after TransformElement?

route-rs-runtime/src/element/transform_elements.rs
eeverett6: I believe so, I copied this from identity_element as well. Update to phantom_in and phantom_out to be more clear.

CollinValley:
Now that Drew's issues have been resolved, this looks good.  You need to make the one change to make clippy happy first though, then we can merge.

CollinValley (approved):
Looks good

PR #85, opened 2019-08-27
afpacket: low-level crate for AF_PACKET

moosingin3space:
This patch adds a new crate implementing low-level safe bindings to
AF_PACKET. The goal of this crate is to create a misuse-resistant
binding to the underlying AF_PACKET library that can be used from safe
Rust.

This crate defines a `Socket` type that is loosely based on the stdlib's
`UdpSocket` type.

### DIFF
--- a/afpacket/src/sockets.rs
+++ b/afpacket/src/sockets.rs
@@ -1,0 +1,212 @@
+#![deny(missing_docs)]
+
+use crate::linux;
+use libc;
+use std::{
+    ffi::CStr,
+    io,
+    mem::{self, MaybeUninit},
+    ptr,
+};
+
+/// Represents a link-local address.
+/// At this time, it's not particularly useful.
+pub struct Addr {
+    _inner: libc::sockaddr_storage,
+    _len: libc::socklen_t,
+}
+
+/// Represents an unbound `AF_PACKET` socket.  At this phase of a socket's lifecycle, it can be
+/// configured.
+pub struct Socket {
+    fd: libc::c_int,
+}
+
+/// Represents a bound `AF_PACKET` socket. At this phase of a socket's lifecycle, it can be read
+/// to/written from.
+pub struct BoundSocket {
+    fd: libc::c_int,
+    send_addr: libc::sockaddr_ll,
+}
+
+impl Socket {
+    /// Creates a new unbound socket.
+    pub fn new() -> io::Result<Self> {
+        // This block must be marked as unsafe because it uses FFI with C code. We believe the code
+        // in this block to be safe because it does not interact with any memory owned by Rust
+        // code, nor does it violate the invariant of the Socket type -- namely, that it return an
+        // Err if it fails to initialize.
+        let fd = unsafe {
+            // Resources:
+            // https://beej.us/guide/bgnet/html/multi/syscalls.html#socket
+            // man 7 packet
+            let fd = libc::socket(libc::AF_PACKET, libc::SOCK_RAW, libc::ETH_P_ALL.to_be());
+            if fd < 0 {
+                return Err(io::Error::last_os_error());
+            }
+            fd
+        };
+        Ok(Self { fd })
+    }
+
+    /// Binds the socket to a network interface. This function consumes the `Socket` instance, as
+    /// no more configuration options may be safely changed.
+    pub fn bind(self, iface: impl AsRef<CStr>) -> io::Result<BoundSocket> {
+        // This block is marked as unsafe because it uses FFI, however, we believe it to be safe
+        // because 1) it handles FFI failures in accordance with the bound API's conventions, and
+        // 2) it safely borrows the &CStr passed in. We will test this functionality with `miri` to
+        // confirm the above assumptions.
+        let send_addr = unsafe {
+            // get the index of the interface
+            let mut ifr: linux::ifreq = MaybeUninit::zeroed().assume_init();
+            ptr::copy_nonoverlapping(
+                iface.as_ref().as_ptr(),
+                ifr.ifr_ifrn.ifrn_name.as_mut_ptr(),
+                libc::IFNAMSIZ,
+            );
+            // ioctl(SIOCGIFINDEX) fills in the index field of the ifreq object
+            // Resources:
+            // man 7 netdevice
+            let err = libc::ioctl(self.fd, linux::SIOCGIFINDEX, &ifr);
+            if err < 0 {
+                return Err(io::Error::last_os_error());
+            }
+
+            // bind the socket
+            let mut ll: libc::sockaddr_ll = MaybeUninit::zeroed().assume_init();
+            ll.sll_family = libc::AF_PACKET as libc::c_ushort;
+            ll.sll_ifindex = ifr.ifr_ifru.ifru_ivalue; // expanded from `ifr_ifindex` in kernel headers
+                                                       // Resources:
+                                                       // https://beej.us/guide/bgnet/html/multi/syscalls.html#bind
+                                                       // man 7 packet regarding sockaddr_ll
+            let err = libc::bind(
+                self.fd,
+                &mut ll as *mut _ as *mut libc::sockaddr,
+                mem::size_of::<libc::sockaddr_ll>() as libc::c_uint,
+            );
+            if err < 0 {
+                return Err(io::Error::last_os_error());
+            }
+            ll
+        };
+        let fd = self.fd;
+        // This ensures that `self` does not attempt to close the file descriptor, as the file
+        // descriptor is transferred to the BoundSocket we're returning. This doesn't cause any
+        // resource leaks since the stack-bound `self` is consumed and deallocated in
+        // `mem::forget`.
+        mem::forget(self);
+        Ok(BoundSocket { fd, send_addr })
+    }
+
+    /// Configures the socket's non-blocking status.
+    pub fn set_nonblocking(&mut self, nonblocking: bool) -> io::Result<()> {
+        // This block is marked as unsafe because it uses FFI, however, we assume this code to be
+        // safe because we handle fcntl's failures properly. Additionally, we do not borrow any
+        // Rust-owned memory.
+        // Resources used to write syscall code:
+        // https://beej.us/guide/bgnet/html/multi/advanced.html#blocking
+        // man 2 fcntl
+        unsafe {
+            let flags = libc::fcntl(self.fd, libc::F_GETFL);
+            if flags < 0 {
+                return Err(io::Error::last_os_error());
+            }
+            let new_flags = if nonblocking {
+                flags | libc::O_NONBLOCK
+            } else {
+                flags & (!libc::O_NONBLOCK)
+            };
+            let err = libc::fcntl(self.fd, libc::F_SETFL, new_flags);
+            if err < 0 {
+                return Err(io::Error::last_os_error());
+            }
+        }
+        Ok(())
+    }
+
+    /// Returns true if the socket is configured not to block, false otherwise.
+    pub fn is_nonblocking(&self) -> io::Result<bool> {
+        // See comments on block above (in set_nonblocking).
+        let flags = unsafe {
+            let flags = libc::fcntl(self.fd, libc::F_GETFL);
+            if flags < 0 {
+                return Err(io::Error::last_os_error());
+            }
+            flags
+        };
+        Ok(flags & libc::O_NONBLOCK == libc::O_NONBLOCK)
+    }
+}
+
+impl BoundSocket {
+    /// Sends a frame to the NIC.
+    pub fn send(&mut self, frame: &[u8]) -> io::Result<usize> {
+        // This block is marked as unsafe because it uses FFI. We believe this code to be safe,
+        // because it safely borrows the Rust-owned frame and passes the length of the frame to the
+        // libc function, so it should not exhibit any C-side undefined behaviour.
+        unsafe {
+            // Resources:
+            // https://beej.us/guide/bgnet/html/multi/syscalls.html#sendtorecv
+            let bytes = libc::sendto(
+                self.fd,
+                frame.as_ptr() as *const _,
+                frame.len(),
+                0,
+                &self.send_addr as *const _ as *const libc::sockaddr,
+                mem::size_of::<libc::sockaddr_ll>() as libc::socklen_t,
+            );
+            if bytes < 0 {
+                Err(io::Error::last_os_error())
+            } else {
+                Ok(bytes as usize)
+            }
+        }
+    }
+
+    /// Receives a frame from the NIC.
+    pub fn recv(&mut self, frame: &mut [u8]) -> io::Result<(usize, Addr)> {
+        // Note comment in `send` call.
+        unsafe {
+            let mut storage = MaybeUninit::<libc::sockaddr_storage>::zeroed();
+            let mut addrlen = mem::size_of::<libc::sockaddr_storage>() as libc::socklen_t;
+
+            // Resources:
+            // https://beej.us/guide/bgnet/html/multi/syscalls.html#sendtorecv
+            let bytes = libc::recvfrom(
+                self.fd,
+                frame.as_mut_ptr() as *mut _,
+                frame.len(),
+                0,
+                storage.as_mut_ptr() as *mut _,
+                &mut addrlen,
+            );
+            if bytes < 0 {
+                Err(io::Error::last_os_error())
+            } else {
+                Ok((
+                    bytes as usize,
+                    Addr {
+                        _inner: storage.assume_init(),
+                        _len: addrlen,
+                    },
+                ))
+            }
+        }
+    }
+}
+
+impl Drop for Socket {
+    fn drop(&mut self) {
+        unsafe {
+            libc::close(self.fd);
+        }
+    }
+}
+
+impl Drop for BoundSocket {
+    fn drop(&mut self) {
+        unsafe {
+            libc::close(self.fd);
+        }
+    }
+}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,6 +5,9 @@ members = [
     "route-rs-graphgen",
     "route-rs-runtime",
 
+    # I/O Crates
+    "afpacket",
+
     # Example crates
     # These crates show usage of route-rs features and should _not_ be published to crates.io
     "examples/trivial-identity",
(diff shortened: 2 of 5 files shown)


ddowl:
Commit nit: "misuse-resistant" is a double negative. Maybe "foolproof" or "The goal of this crate is to create a simple interface to the underlying `AF_PACKET` ..."

moosingin3space:
The statement "misuse-resistant" is used a lot in cryptography, for instance [here](https://tonyarcieri.com/introducing-miscreant-a-multi-language-misuse-resistant-encryption-library).

ddowl:
Initially glancing at `Socket` it seems to have the interface I'd expect, basically `bind`, `send`, and `recv`, with a modifier to set nonblocking. I'm curious why there isn't a `listen` as the complement to `bind`, and why there's a nonblocking modifier as opposed to `send/try_send` and `recv/try_recv` variants like we see on channels? 

Also, as a reviewer that's not super familiar with this kind of low-level unsafe Rust, it might be nice to see examples, or references of how you came up with your code.

Also tests. That would give me confidence that your stuff probably works as intended, even if I'm not super familiar with the code.

moosingin3space:
* `listen` isn't applicable to `AF_PACKET` sockets -- since there's no real concept of a server or client. Likewise, there's no `connect` operation.
* Non-blocking has to be set on the socket, not on the operation. That's a sockets API thing.
* By examples or references, do you mean C code that it's the equivalent of? `man` pages I used? Or would something like [Beej's Guide to Network Programming](https://beej.us/guide/bgnet/html/multi/index.html) be sufficient?
* Testing this is a PITA because it requires root (and can't run in a container without adding caps). I love tests and would definitely want to have a better testing story around this. @scgruber can you tell if Travis CI supports this? If so, I've got some ideas for testing this.

ddowl:
>     * By examples or references, do you mean C code that it's the equivalent of? `man` pages I used? Or would something like [Beej's Guide to Network Programming](https://beej.us/guide/bgnet/html/multi/index.html) be sufficient?

The man pages and this network programming guide would be perfect.

>     * Testing this is a PITA because it requires root (and can't run in a container without adding caps). I love tests and would definitely want to have a better testing story around this. @scgruber can you tell if Travis CI supports this? If so, I've got some ideas for testing this.

I wonder if there's a way to make the logic more testable. For example, injecting dependencies that would use root, but then supply some kind of stub in a testing environment. Stubbing too much starts to ruin the purpose of the test at some point, so 🤷‍♂

moosingin3space:
I guess we could mock out the `libc` crate and `linux` module with a strategy similar to [this](https://klausi.github.io/rustnish/2019/03/31/mocking-in-rust-with-conditional-compilation.html). In that case, it might make sense to make the `linux` module a facade over `libc` so we're only mocking out one thing.

I still want integration tests that would test this code against a real Linux kernel -- and for that we'd need root.

moosingin3space:
Another thing I just realized is that we should probably try to use miri or cargo-fuzz on our mocked tests (if we go that route). Since this crate will (hopefully) concentrate all our unsafe usage, we should test it as much as possible!

afpacket/src/sockets.rs
```diff
@@ -0,0 +1,181 @@
...
+/// to/written from.
+pub struct BoundSocket {
+    fd: libc::c_int,
+    send_addr: libc::sockaddr_ll,
+}
+
+impl Socket {
+    /// Creates a new unbound socket.
+    pub fn new() -> io::Result<Self> {
+        // This block is unsafe because it invokes a syscall. It is safe because it checks the
```
ddowl: Is unsafe and is safe? Do you mean that the block below is unsafe but the `new` fn is safe?

afpacket/src/sockets.rs
```diff
@@ -0,0 +1,181 @@
...
+        // syscall return value and returns an Err if the sycall fails. It does not interact with
+        // Rust-owned memory, so it's safe as per the borrow checker's rules.
+        let fd = unsafe {
+            // Resources:
+            // https://beej.us/guide/bgnet/html/multi/syscalls.html#socket
+            // man 7 packet
+            let fd = libc::socket(libc::AF_PACKET, libc::SOCK_RAW, libc::ETH_P_ALL.to_be());
```
ddowl, line 43: "to be or not to be"

ddowl: Why are we manipulating endianness of `ETH_P_ALL`? I also couldn't really figure out what that meant. Is it "give me all the packets regardless of packet type"? [Relevant search](https://docs.rs/libc/0.2.54/libc/constant.ETH_P_ALL.html?search=eth_p)

afpacket/src/sockets.rs
moosingin3space: Sorry, this was a really poorly-worded comment. The idea I'm trying to communicate here is that while I'm using an `unsafe` block, the code in the `unsafe` block doesn't violate the borrow checker's invariants nor does it violate the type-system-enforced invariants of the `Socket` struct.

afpacket/src/sockets.rs
moosingin3space: > When protocol is set to htons(ETH_P_ALL), then all protocols are received.  All incoming packets of that protocol type will be passed to the packet socket before they are passed to the protocols implemented in the kernel.

[source](http://man7.org/linux/man-pages/man7/packet.7.html).

`htons` in C is "host to network byte order for `short`" (i.e. convert unsigned 16-bit integer to big-endian, which is network byte order). In Rust, that's `to_be`.

afpacket/src/sockets.rs
ddowl: Maybe we can communicate that with something like "The block below is mostly safe because A, B, C, but we use `unsafe` since we're making a `socket` syscall"?

afpacket/src/sockets.rs
ddowl: Neat! TIL "network byte order" is big-endian.

afpacket/src/sockets.rs
```diff
@@ -0,0 +1,181 @@
...
+            }
+            fd
+        };
+        Ok(Self { fd })
+    }
+
+    /// Binds the socket to a network interface. This function consumes the `Socket` instance, as
+    /// no more configuration options may be safely changed.
+    pub fn bind(self, iface: impl AsRef<CStr>) -> io::Result<BoundSocket> {
+        // This block is unsafe because it invokes syscalls. It is safe because the syscall return
```
ddowl: Same wording as the comment above. "This call is mostly safe since the syscall return values are checked and `Err` is returned if necessary, but we're still making syscalls".

afpacket/src/sockets.rs
```diff
@@ -0,0 +1,181 @@
...
+                return Err(io::Error::last_os_error())
+            }
+            fd
+        };
+        Ok(Self { fd })
+    }
+
+    /// Binds the socket to a network interface. This function consumes the `Socket` instance, as
+    /// no more configuration options may be safely changed.
+    pub fn bind(self, iface: impl AsRef<CStr>) -> io::Result<BoundSocket> {
```
ddowl, line 54: The interface to `bind` takes a `sockaddr` as a descriptor to where to bind the given `Socket`. `sockaddr` is [essentially](https://beej.us/guide/bgnet/html/multi/sockaddr_inman.html) a struct with IP and port, so why does our `Socket#bind` take a C string? Is there a reason we can't bind the type more strictly?

afpacket/src/sockets.rs
```diff
@@ -0,0 +1,181 @@
...
+
+            // bind the socket
+            let mut ll : libc::sockaddr_ll = MaybeUninit::zeroed().assume_init();
+            ll.sll_family = libc::AF_PACKET as libc::c_ushort;
+            ll.sll_ifindex = ifr.ifr_ifru.ifru_ivalue; // expanded from `ifr_ifindex` in kernel headers
+            // Resources:
+            // https://beej.us/guide/bgnet/html/multi/syscalls.html#bind
+            // man 7 packet regarding sockaddr_ll
+            let err = libc::bind(self.fd,
+                                 &mut ll as *mut _ as *mut libc::sockaddr,
```
ddowl: `as *mut _ as *mut libc::sockaddr` excuse me wat?

afpacket/src/sockets.rs
```diff
@@ -0,0 +1,181 @@
...
+            // man 7 netdevice
+            let err = libc::ioctl(self.fd, linux::SIOCGIFINDEX, &ifr);
+            if err < 0 {
+                return Err(io::Error::last_os_error())
+            }
+
+            // bind the socket
+            let mut ll : libc::sockaddr_ll = MaybeUninit::zeroed().assume_init();
+            ll.sll_family = libc::AF_PACKET as libc::c_ushort;
+            ll.sll_ifindex = ifr.ifr_ifru.ifru_ivalue; // expanded from `ifr_ifindex` in kernel headers
```
ddowl, line 78: To paraphrase for my own understanding, we use `ioctl` to determine the _physical_ interface of the thing we're trying to bind to, and here we're setting that on the actual `sockaddr_ll` that we're binding to. I feel like I'm forced to conclude that this some representation of a port number, since there's no other information transferred from the `iface` argument right?

Is there a reason we aren't setting the other fields on the `sockaddr_ll` struct?

afpacket/src/sockets.rs
```diff
@@ -0,0 +1,181 @@
...
+            let err = libc::bind(self.fd,
+                                 &mut ll as *mut _ as *mut libc::sockaddr,
+                                 mem::size_of::<libc::sockaddr_ll>() as libc::c_uint);
+            if err < 0 {
+                return Err(io::Error::last_os_error())
+            }
+            ll
+        };
+        let fd = self.fd;
+        mem::forget(self);
```
ddowl, line 97: This is a little confusing to me. Since we want to use the same `fd` for a `BoundSocket` we run `mem::forget(self)`, essentially leaking/disposing the memory without running `Socket`'s destructor. So what happens if I drop a `Socket` after creating a `BoundSocket` from it? Wouldn't that just drop the same `fd`? I feel like there might be a better way to model this state transition, or maybe I'm misinterpreting this behavior?

afpacket/src/sockets.rs
```diff
@@ -0,0 +1,181 @@
...
+        Ok(BoundSocket { fd, send_addr })
+    }
+
+    /// Configures the socket's non-blocking status.
+    pub fn set_nonblocking(&mut self, nonblocking: bool) -> io::Result<()> {
+        // This block is unsafe because it requires syscalls, but it is safe because each syscall's
```
ddowl: Ditto comment wording.

ddowl:
Wanted to mention that including inline comments to docs for particular syscalls and low-level networking stuff has been really helpful as I review and attempt to understand this code. Kudos @moosingin3space !

afpacket/src/sockets.rs
moosingin3space: The first part `&mut ll as *mut _` converts `&mut ll` from `&mut libc::sockaddr_ll` to `*mut libc::sockaddr_ll`. After we've converted it to a raw pointer, we then cast the raw pointer to the right type for the function (`*mut libc::sockaddr`).

afpacket/src/sockets.rs
moosingin3space: The idea is that this function *consumes* the `Socket`, but hands its `fd` to a `BoundSocket`. So now the `BoundSocket` is responsible for closing the `fd`.

Without calling `mem::forget(self)`, `self` (the `Socket`) has its destructor run at the end of `bind`, which would close the `fd` before client code uses it in `BoundSocket`.

afpacket/src/sockets.rs
```diff
@@ -0,0 +1,181 @@
...
+            }
+            flags
+        };
+        Ok(flags & libc::O_NONBLOCK == libc::O_NONBLOCK)
+    }
+}
+
+impl BoundSocket {
+    /// Sends a frame to the NIC.
+    pub fn send(&mut self, frame: &[u8]) -> io::Result<usize> {
```
ddowl, line 143: Is the idea here that clients of `BoundSocket` will just serialize whatever data type they're working with to a slice of bytes to be sent? I like the interface since it's the most flexible, but it means we should include serialization helpers in the `smoltcp` types we're using in `runtime` if that's necessary.

afpacket/src/sockets.rs
moosingin3space: Beej's guide doesn't really cover `AF_PACKET` sockets, since it's about stuff that's common to all Unix OSes. Again, noting the [man page](http://man7.org/linux/man-pages/man7/packet.7.html), it takes a `sockaddr_ll` struct, which represents the network interface. That's why I'm taking a `&CStr` -- it's supposed to represent the name of the network interface the caller wants to bind this socket to.

We could define a `NetworkInterface` struct and pass that in instead of a `&CStr`.

afpacket/src/sockets.rs
ddowl: I answered my second question. `bind` populates the `sockaddr_ll` and is given to a new `BoundSocket`.

afpacket/src/sockets.rs
```diff
@@ -0,0 +1,181 @@
...
+        unsafe {
+            // Resources:
+            // https://beej.us/guide/bgnet/html/multi/syscalls.html#sendtorecv
+            let bytes = libc::sendto(self.fd,
```
ddowl: > Just like with `send()`, `sendto()` returns the number of bytes actually sent (which, again, might be less than the number of bytes you told it to send!), or -1 on error.

Should we loop until the whole frame is sent?

afpacket/src/sockets.rs
ddowl: Could we also represent the state machine transition by having a `send_addr: Option<libc::sockaddr_ll>,` in `Socket`? We wouldn't need to deal with this memory management stuff, but having `Socket` and `BoundSocket` distinct really makes it hard to use the API incorrectly (e.g. you can't `send` on a socket that isn't bound.). I'm not sure what the right design here is.

moosingin3space: That's what my original design did, but it made it possible to call `send` on an unbound socket, which is not ideal.

afpacket/src/sockets.rs
moosingin3space: I don't think so. No other socket implementations in Rust do that.

afpacket/src/sockets.rs
moosingin3space: Should we resolve this then?

afpacket/src/sockets.rs
ddowl: Yeah a `NetworkInterface` would be preferable.

afpacket/src/sockets.rs
ddowl: If my guess that `ifindex` represents port number, then yes!

afpacket/src/sockets.rs
ddowl: I guess I've never seen a reference casted to a raw pointer before. Strange that we can't do `&mut ll as *mut libc::sockaddr`.

afpacket/src/sockets.rs
ddowl: Oh I see. There's no way to drop the consumed `Socket` here. The `mem::forget` documentation is a little scary:

> Any resources the value manages, such as heap memory or a file handle, will linger forever in an unreachable state.

It would be nice to know that the memory is actually being released properly without closing the file descriptor...

afpacket/src/sockets.rs
ddowl: Oh cool, we return back the bytes read in the `Result` so the client can do that. SGTM

afpacket/src/sockets.rs
ddowl: The other options I can think of are

1. Add a `close` method and require the client to explicitly call it to close the file descriptor.
2. Bind to a different fd for `BoundSocket` and drop the first one.
3. Convince me that the `mem::forget` option isn't as bad as I think.

afpacket/src/sockets.rs
moosingin3space: It's not really a "port number" in the same sense as a TCP/UDP socket. It should be thought of as the NIC's index.

For instance:
```
moosnat@asimov:~$ ip link
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
2: enp4s0: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc fq_codel state DOWN mode DEFAULT group default qlen 1000
    link/ether e0:3f:49:2c:3d:d5 brd ff:ff:ff:ff:ff:ff
3: wlp3s0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP mode DORMANT group default qlen 1000
    link/ether 54:27:1e:07:1b:cd brd ff:ff:ff:ff:ff:ff
```

The NIC index is displayed in the output.

afpacket/src/sockets.rs
moosingin3space: The `mem::forget` option is safe, because the `Socket` object manages the file descriptor. That file descriptor was moved to `BoundSocket`, which enables it to handle the lifecycle of the file descriptor.

afpacket/src/sockets.rs
moosingin3space: Yes. Just serialize the data into a slice. I'd rather not depend on `smoltcp` here.

afpacket/src/sockets.rs
moosingin3space: Just wanted to follow up that only the following casts are valid (explaining why we need two casts):
* `&T as *const T`
* `&mut T as *mut T`
* `*const T as *const U`
* `*mut T as *mut U`
* `*mut T as *const T`

afpacket/src/sockets.rs
moosingin3space: What sort of data should this `NetworkInterface` expose?
Aside, I found [this](http://man7.org/linux/man-pages/man3/if_nametoindex.3.html), which could simplify the `ifreq` dance.

ddowl: Discussed offline, the network interface here is the _NIC_ name, which doesn't have a well-defined structure at compile-time.

ddowl (dismissed):
LGTMMD

ddowl (dismissed):
LGTMMD

ddowl:
resolves #19

ddowl (approved):
_LGTMMD_

PR #86, opened 2019-08-27
Add Tee link element

ErrHu:
This is basically a copy of Classifier except the processing
element is replaced with a loop that clones the packet.

I would like comments on 
tee_link.rs:107
where I only clone the packet. Is there a better way to clone until the last index then consume?
Also I think I re-encountered an issue where is not all branches are pulled from, the whole system will stall. I think we should check for this scenario (making it another issue).

### DIFF
--- a/route-rs-runtime/src/link/tee_link.rs
+++ b/route-rs-runtime/src/link/tee_link.rs
@@ -1,0 +1,333 @@
+use crate::link::task_park::*;
+use crate::link::PacketStream;
+use crossbeam::atomic::AtomicCell;
+use crossbeam::crossbeam_channel;
+use crossbeam::crossbeam_channel::{Receiver, Sender, TryRecvError};
+use futures::{Async, Future, Poll, Stream};
+use std::sync::Arc;
+
+pub struct TeeLink<P: Sized + Clone> {
+    pub ingressor: TeeIngressor<P>,
+    pub egressors: Vec<TeeEgressor<P>>,
+}
+
+impl<P: Sized + Clone> TeeLink<P> {
+    pub fn new(input_stream: PacketStream<P>, queue_capacity: usize, branches: usize) -> Self {
+        assert!(
+            branches <= 1000,
+            format!("Tee Link branches: {} > 1000", branches)
+        );
+        assert!(
+            queue_capacity <= 1000,
+            format!("Tee Link queue_capacity: {} > 1000", queue_capacity)
+        );
+
+        let mut to_egressors: Vec<Sender<Option<P>>> = Vec::new();
+        let mut egressors: Vec<TeeEgressor<P>> = Vec::new();
+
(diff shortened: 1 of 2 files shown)


route-rs-runtime/src/link/tee_link.rs
```diff
@@ -0,0 +1,336 @@
+use crate::link::task_park::*;
+use crate::link::PacketStream;
+use crossbeam::atomic::AtomicCell;
+use crossbeam::crossbeam_channel;
+use crossbeam::crossbeam_channel::{Receiver, Sender, TryRecvError};
+use futures::{Async, Future, Poll, Stream};
+use std::sync::Arc;
+
+pub struct TeeElementLink<P: Sized + Clone> {
+    pub consumer: TeeElementConsumer<P>,
```
CollinValley: Going to need to switch this to the new naming nomenclature (ingressor, egressor).

route-rs-runtime/src/link/tee_link.rs
```diff
@@ -0,0 +1,336 @@
...
+    pub providers: Vec<TeeElementProvider<P>>,
+}
+
+impl<P: Sized + Clone> TeeElementLink<P> {
```
CollinValley: Rename TeeElementLink to TeeLink, same concept for other sub-structures.

route-rs-runtime/src/link/tee_link.rs
```diff
@@ -0,0 +1,336 @@
...
+                if to_provider.is_full() {
+                    park_and_notify(&self.task_parks[port]);
+                    return Ok(Async::NotReady);
+                }
+            }
+            let packet_option: Option<P> = try_ready!(self.input_stream.poll());
+
+            match packet_option {
+                None => return Ok(Async::Ready(())),
+                Some(packet) => {
```
CollinValley, line 103: This is fine to just the clone the packet each time.  The original copy will just be dropped when it goes out of scope, when the poll function ends. You could be slightly more efficient, by iterating over `0..self.task_parks.len()-1`,  and then doing a non-clone send into the last slot.  But honestly, that makes the code double the verbosity without much gain. I doubt the performance changes at all.

route-rs-runtime/src/link/tee_link.rs
```diff
@@ -0,0 +1,336 @@
...
+    }
+}
+
+impl<P: Sized + Clone> Future for TeeElementConsumer<P> {
+    type Item = ();
+    type Error = ();
+
+    /// Same logic as AsyncElementConsumer, except if any of the channels are full we
+    /// await that channel to clear before processing a new packet. This is somewhat
+    /// inefficient, but seems acceptable for now since we want to yield compute to
```
CollinValley: The second sentence of this comment doesn't make sense in this context.

CollinValley (approved):
This looks good, there may be some places for future improvement, can do stuff like that later.

Issue #14, opened 2019-07-18
Create "basic transformation elements"


ddowl:
This is referring to the elements provided in the [Click paper](http://www.read.seas.harvard.edu/~kohler/z/kohler00click-yapteaparprfotci.pdf). 

These are:

1. **ARPQuerier(...)** Push. First input takes IP packets, second input takes ARP responses with Ethernet headers. Output emits ARP queries and IP-inEthernet packets. Uses ARP to find the Ethernet address corresponding to each input IP packet’s destination IP address annotation; encapsulates the packet in an Ethernet header with that destination Ethernet address.
1. **ARPResponder(ip eth, ...)** Agnostic. Input takes ARP queries, output emits ARP responses. Responds to ARP queries for IP address ip with the static Ethernet address eth.
1. **CheckIPHeader(...)** Agnostic. Input takes IP packets. Discards packets with invalid IP length, source address, or checksum fields; forwards valid packets unchanged.
1. **CheckPaint(p)** Agnostic. Input takes any packet. Forwards packets with paint annotation p to both outputs; otherwise just to first.
1. **Classifier(...)** Push. Input takes any packet. Examines packet data according to a set of classifiers, one classifier per output port. Forwards packet to output port corresponding to the first classifier that matched. Example classifier: “12/0800” checks that the packet’s data bytes 12 and 13 contain values 8 and 0, respectively. See also IPClassifier.
1. **DecIPTTL** Agnostic. Input takes IP packets. Decrements input packets’ IP time-to-live field. If the packet is still live, incrementally updates IP checksum and sends modified packet to first output; if it  has expired, sends unmodified packet to second output.
1. **Discard** Push. Discards all input packets.
1. **DropBroadcasts** Agnostic. Input takes any packet. Discards packets that arrived as link-level broadcasts; forwards others unchanged.
1. **EtherSpanTree** Agnostic. Inputs take IEEE 802.1d messages. Implements the IEEE 802.1d spanning tree algorithm for Ethernet switches.
1. **EtherSwitch** Push. Inputs take Ethernet packets; one input per output. Learning, forwarding Ethernet switch. Learns the interfaces corresponding to Ethernet addresses by examining input packets’ source addresses; forwards packets to the correct output port, or broadcasts them if the destination Ethernet address is not yet known.
[...]

PR #87, opened 2019-08-29
Add DecIpv4Hop and DecIpv6Hop Elements

eeverett6:
Adds an element to decrement the TTL of an IPv4 packet. Adds a
dependency on smoltcp for packet processing.

This is one of the basic transformation elements from the Click paper,
see #14


eeverett6:
Wanted to get a trivial networking element in to make sure we're all on the same page. I'm not sure this belongs in the runtime, but I know it's one of the basic elements we'll want to provide.

The warning generated in smoltcp seems to be referenced here, but nowhere else I can find: https://github.com/m-labs/smoltcp/issues/242

CollinValley:
This is basically exactly what I was thinking.  I think we are on the same page here. This could become an element on it's own if you want to resubmit the PR,.  I'm not enamored with the name DecIPTTLElement,  but renaming is easy to do later too.

eeverett6:
@scgruber just want to check with you as well before adding 0BSD as a compatible license for our dependencies.

route-rs-runtime/src/element/dec_ip_ttl.rs
```diff
@@ -0,0 +1,52 @@
+use crate::element::Element;
+use smoltcp::wire::*;
+
+/*
+  DecIPTTL
+  Decrements the TTL of an IPv4 packet
+*/
+#[derive(Default)]
+pub struct DecIPTTLElement {}
```
moosingin3space: Rust conventions are to avoid having a lot of capital letters next to each other in an identifier. For instance, `DecIpv4HopLimit` is probably a better name. (This is to avoid inconsistency like Java's `URLConnection` and `HttpURLConnection`.)

route-rs-runtime/Cargo.toml
```diff
@@ -8,4 +8,5 @@ license = "MIT"
 [dependencies]
 tokio = "*"
 futures = "*"
-crossbeam = "*"
\ No newline at end of file
+crossbeam = "*"
+smoltcp = "*"
```
moosingin3space: We should probably pin the version of `smoltcp`.

route-rs-runtime/src/element/dec_ip_ttl.rs
eeverett6: Is DecIpv4Hop sufficient? DecIpv4HopLimitElement just feels a bit wordy

moosingin3space: How about no `Element` suffix? It's in the `element` module, we don't need to redundantly call it an `Element`.

eeverett6: Makes sense, updated

route-rs-runtime/Cargo.toml
CollinValley: Good idea, especially since we are only using this for its internal structures.  Once async/await lands and we port tokio I plan on pinning all our dependencies, and maybe make a monthly issue to try and bump the versions when we can.

CollinValley (dismissed):
I approve these changes, but I would like to get a :+1:  from @scgruber about adding the new acceptable license. This one seems compatible with our license schema from my reading but I'd like another pass from someone more versed.

route-rs-runtime/src/element/dec_ip_hop.rs
```diff
@@ -0,0 +1,96 @@
+use crate::element::Element;
+use smoltcp::wire::*;
+
+/*
+  DecIpv4Hop
+  Decrements the TTL of an IPv4 packet
```
scgruber: Conventionally, we should use `///` comments here so they show in documentation. We also don't need to repeat the name in the header since it is shown immediately below.

deny.toml
```diff
@@ -4,6 +4,7 @@ unlicensed = "deny"
 # This lists all of the licenses we will allow in our dependencies. Since route-rs is distributed
 # under the MIT license, we want to make sure all of our dependencies are similarly unrestrictive.
 allow = [
+    "0BSD",
```
scgruber, line 7: LGTM

route-rs-runtime/Cargo.toml
```diff
@@ -8,4 +8,5 @@ license = "MIT"
 [dependencies]
 tokio = "*"
 futures = "*"
-crossbeam = "*"
\ No newline at end of file
+crossbeam = "*"
+smoltcp = "0.5.0"
```
scgruber, line 12: For now, let's let our dependencies float unless we have a good reason to lock them.

route-rs-runtime/src/element/dec_ip_hop.rs
```diff
@@ -0,0 +1,96 @@
...
+        DecIpv4HopLimit {}
+    }
+}
+
+impl Element for DecIpv4HopLimit {
+    type Input = Ipv4Packet<Vec<u8>>;
+    type Output = Ipv4Packet<Vec<u8>>;
+
+    fn process(&mut self, mut packet: Self::Input) -> Self::Output {
+        packet.set_hop_limit(packet.hop_limit() - 1);
```
scgruber: What happens if the hop limit is currently the minimum allowed? Can we generate an invalid packet by doing this?

route-rs-runtime/src/element/dec_ip_hop.rs
```diff
@@ -0,0 +1,96 @@
...
+        packet
+    }
+}
+
+/*
+  DecIpv6Hop
+  Decrements the TTL of an IPv6 packet
```
scgruber: Same as above about comment formatting

route-rs-runtime/src/element/dec_ip_hop.rs
```diff
@@ -0,0 +1,96 @@
...
+        DecIpv6HopLimit {}
+    }
+}
+
+impl Element for DecIpv6HopLimit {
+    type Input = Ipv6Packet<Vec<u8>>;
+    type Output = Ipv6Packet<Vec<u8>>;
+
+    fn process(&mut self, mut packet: Self::Input) -> Self::Output {
+        packet.set_hop_limit(packet.hop_limit() - 1);
```
scgruber: Same as above about minimum values

route-rs-runtime/src/element/dec_ip_hop.rs
```diff
@@ -0,0 +1,96 @@
...
+        packet
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
scgruber, line 55: You should add tests for what happens at hop_limit = 0

route-rs-runtime/src/element/mod.rs
```diff
@@ -6,3 +6,6 @@ pub use self::identity_elements::*;
 
 mod transform_elements;
 pub use self::transform_elements::*;
+
+mod dec_ip_hop;
```
scgruber, line 10: I do like this general structure of putting the IPv4 and IPv6 elements in the same file. It would be even nicer if we could somehow generate both from the same specification, but that's definitely future enhancements, not POC.

scgruber (changes requested):

route-rs-runtime/src/element/dec_ip_hop.rs
CollinValley: Seconded, we should begin migrating to docgen comments when it makes sense

route-rs-runtime/src/element/dec_ip_hop.rs
CollinValley: One quick fix is to just never set this value below zero, and then leave it up to a classify element later to check ttl is greater than zero? The click codebase will be a decent guide

route-rs-runtime/src/element/dec_ip_hop.rs
eeverett6: Done

route-rs-runtime/src/element/dec_ip_hop.rs
eeverett6: I'm limiting this to 0 and passing the packet along unchanged. When we have more support for ICMP errors in the future we can generate an ICMP error instead.

scgruber (approved):

PR #88, opened 2019-08-29
Replace simple 'panic' on unwrap with 'expect'

ErrHu:
There are many places where a Option is checked/unwrapped and then
panics if None. This can be replaced with expect.

There are some more complicated panics that print other variables.
This would require a closure to generate the string if using an
expect. Will consider later.

### DIFF
--- a/route-rs-runtime/src/link/async_link.rs
+++ b/route-rs-runtime/src/link/async_link.rs
@@ -82,9 +82,9 @@ impl<E: AsyncElement> AsyncIngressor<E> {
 /// it in the future.
 impl<E: AsyncElement> Drop for AsyncIngressor<E> {
     fn drop(&mut self) {
-        if let Err(err) = self.to_egressor.try_send(None) {
-            panic!("Ingressor: Drop: try_send to_Egressor, fail?: {:?}", err);
-        }
+        self.to_egressor
+            .try_send(None)
+            .expect("AsyncIngressor::Drop: try_send to_egressor shouldn't fail");
         die_and_notify(&self.task_park);
     }
 }
@@ -122,12 +122,9 @@ impl<E: AsyncElement> Future for AsyncIngressor<E> {
                 None => return Ok(Async::Ready(())),
                 Some(input_packet) => {
                     let output_packet: E::Output = self.element.process(input_packet);
-                    if let Err(err) = self.to_egressor.try_send(Some(output_packet)) {
-                        panic!(
-                            "Error in to_egressor sender, have nowhere to put packet: {:?}",
-                            err
-                        );
-                    }
+                    self.to_egressor
+                        .try_send(Some(output_packet))
+                        .expect("AsyncIngressor::Poll: try_send to_egressor shouldn't fail");
                     unpark_and_notify(&self.task_park);
                 }
             }
--- a/route-rs-runtime/src/link/classify_link.rs
+++ b/route-rs-runtime/src/link/classify_link.rs
@@ -83,11 +83,10 @@ impl<E: ClassifyElement> Drop for ClassifyIngressor<E> {
     fn drop(&mut self) {
         //TODO: do this with a closure or something, this could be a one-liner
         for to_egressor in self.to_egressors.iter() {
-            if let Err(err) = to_egressor.try_send(None) {
-                panic!("ingressor: Drop: try_send to provider, fail?: {:?}", err);
-            }
+            to_egressor
+                .try_send(None)
+                .expect("ClassifyIngressor::Drop: try_send to_egressor shouldn't fail");
         }
-
         for task_park in self.task_parks.iter() {
             die_and_notify(&task_park);
         }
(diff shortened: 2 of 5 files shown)


route-rs-runtime/src/link/async_link.rs
```diff
@@ -82,9 +82,9 @@ impl<E: AsyncElement> AsyncIngressor<E> {
 /// it in the future.
 impl<E: AsyncElement> Drop for AsyncIngressor<E> {
     fn drop(&mut self) {
-        if let Err(err) = self.to_egressor.try_send(None) {
-            panic!("Ingressor: Drop: try_send to_Egressor, fail?: {:?}", err);
-        }
+        self.to_egressor
+            .try_send(None)
+            .expect("Consumer: Drop: try_send to_egressor, fail?");
```
CollinValley: Nit: Should say Ingressor instead of Consumer.
Let's take this momement to say "AsyncLink::Ingressor::Drop:  try_send to_egressor shouldn't fail:" as our error message.  That provides clear indication what function actually sploded.

CollinValley: Actually we can get away with AsyncIngressor::Drop, since that is unique.

route-rs-runtime/src/link/async_link.rs
```diff
@@ -122,12 +122,9 @@ impl<E: AsyncElement> Future for AsyncIngressor<E> {
...
                     let output_packet: E::Output = self.element.process(input_packet);
-                    if let Err(err) = self.to_egressor.try_send(Some(output_packet)) {
-                        panic!(
-                            "Error in to_egressor sender, have nowhere to put packet: {:?}",
-                            err
-                        );
-                    }
+                    self.to_egressor
+                        .try_send(Some(output_packet))
+                        .expect("Error in to_egressor sender, have nowhere to put packet");
```
CollinValley: Same concept as before, let's have "AsyncIngressor::Poll: try_send to_egressor shouldn't  fail:"

route-rs-runtime/src/link/classify_link.rs
```diff
@@ -83,11 +83,10 @@ impl<E: ClassifyElement> Drop for ClassifyIngressor<E> {
     fn drop(&mut self) {
         //TODO: do this with a closure or something, this could be a one-liner
         for to_egressor in self.to_egressors.iter() {
-            if let Err(err) = to_egressor.try_send(None) {
-                panic!("ingressor: Drop: try_send to provider, fail?: {:?}", err);
-            }
+            to_egressor
+                .try_send(None)
+                .expect("Consumer: Drop: try_send to provider, fail?");
```
CollinValley: Nit: Replace consumer with same concept as comments above.

route-rs-runtime/src/link/join_link.rs
```diff
@@ -72,9 +72,9 @@ impl<Packet: Sized> JoinIngressor<Packet> {
 
 impl<Packet: Sized> Drop for JoinIngressor<Packet> {
     fn drop(&mut self) {
-        if let Err(err) = self.to_egressor.try_send(None) {
-            panic!("Ingressor: Drop: try_send to_egressor, fail?: {:?}", err);
-        }
+        self.to_egressor
+            .try_send(None)
+            .expect("Ingressor: Drop: try_send to_egressor, fail?");
```
CollinValley: Same idea.

route-rs-runtime/src/link/join_link.rs
```diff
@@ -111,12 +111,9 @@ impl<Packet: Sized> Future for JoinIngressor<Packet> {
...
                 Some(packet) => {
-                    if let Err(err) = self.to_egressor.try_send(Some(packet)) {
-                        panic!(
-                            "Error in to_egressor sender, have nowhere to put packet: {:?}",
-                            err
-                        );
-                    }
+                    self.to_egressor
+                        .try_send(Some(packet))
+                        .expect("Error in to_egressor sender, have nowhere to put packet");
```
CollinValley: Same

route-rs-runtime/src/pipeline/output_channel_link.rs
```diff
@@ -29,11 +29,10 @@ impl<Output> Future for OutputChannelLink<Output> {
...
             match try_ready!(self.input_stream.poll()) {
-                Some(packet) => {
-                    if let Err(err) = self.output_channel.try_send(packet) {
-                        panic!("OutputChannelLink: Error sending to channel: {:?}", err);
-                    }
-                }
+                Some(packet) => self
+                    .output_channel
+                    .try_send(packet)
+                    .expect("OutputChannelLink: Error sending to channel"),
```
CollinValley: Same,  just add the poll context

CollinValley (commented):
This is a good catch Eric, makes the code quite a bit cleaner. Besides the nits this is ready for merge.

CollinValley (approved):
Everything looks good, thanks for this PR!

PR #89, opened 2019-08-31
Link: Add BlackHole

CollinValley:
This is a simple link that  takes in an input stream and drops any
packets that it recieves. Much like the async_link, it will continue
to pull packets until it receives a NotReady, in which case it will
propogate the NotReady until there are further packets to blackhole.

In general this link will always be handed directly to the runtime.

### DIFF
--- a/route-rs-runtime/src/link/blackhole_link.rs
+++ b/route-rs-runtime/src/link/blackhole_link.rs
@@ -1,0 +1,119 @@
+use crate::link::PacketStream;
+use futures::{Async, Future, Poll};
+
+/// Link that drops all packets ingressed.
+pub struct BlackHoleLink<Packet: Sized> {
+    input_stream: PacketStream<Packet>,
+}
+
+impl<Packet: Sized> BlackHoleLink<Packet> {
+    pub fn new(input_stream: PacketStream<Packet>) -> Self {
+        BlackHoleLink { input_stream }
+    }
+}
+
+impl<Packet: Sized> Future for BlackHoleLink<Packet> {
+    type Item = ();
+    type Error = ();
+
+    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+        loop {
+            let input_packet_option: Option<Packet> = try_ready!(self.input_stream.poll());
+            match input_packet_option {
+                None => return Ok(Async::Ready(())),
+                Some(_) => {}
+            };
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::element::Classifier;
+    use crate::link::ClassifyLink;
+    use crate::utils::test::packet_collectors::ExhaustiveCollector;
+    use crate::utils::test::packet_generators::{immediate_stream, PacketIntervalGenerator};
+    use core::time;
+    use crossbeam::crossbeam_channel;
+    use futures::future::lazy;
+
+    struct ClassifyEvenness {}
+
+    impl ClassifyEvenness {
+        pub fn new() -> Self {
+            ClassifyEvenness {}
+        }
+    }
+
+    impl Classifier for ClassifyEvenness {
+        type Packet = i32;
+        type Class = bool;
+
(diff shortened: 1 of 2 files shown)


route-rs-runtime/src/link/blackhole_link.rs
```diff
@@ -0,0 +1,128 @@
...
+        BlackHoleLink { input_stream }
+    }
+}
+
+impl<Packet: Sized> Future for BlackHoleLink<Packet> {
+    type Item = ();
+    type Error = ();
+
+    /*
+    4 cases: Async::Ready(Some), Async::Ready(None), Async::NotReady, Err
```
scgruber: Use `///` so they are doc comments

route-rs-runtime/src/link/blackhole_link.rs
```diff
@@ -0,0 +1,128 @@
...
+    impl ClassifyElement for ClassifyEvenOddElement {
+        type Packet = i32;
+
+        fn classify(&mut self, packet: &Self::Packet) -> usize {
+            (packet % 2) as usize
+        }
+    }
+
+    #[test]
+    fn one_blackhole_link() {
```
scgruber: Maybe call this `blackhole_link_finishes`

route-rs-runtime/src/link/blackhole_link.rs
```diff
@@ -0,0 +1,128 @@
...
+
+        let link1 = BlackHoleLink::new(Box::new(packet_generator));
+
+        tokio::run(link1);
+
+        //In this test, we just ensure that it finishes.
+    }
+
+    #[test]
+    fn one_blackhole_link_wait_between_packets() {
```
scgruber: Maybe call this `blackhole_link_with_wait_finishes`

route-rs-runtime/src/link/blackhole_link.rs
ddowl: Since we're familiar with the `poll` contract now, these comments probably aren't necessary anymore.

route-rs-runtime/src/link/blackhole_link.rs
```diff
@@ -0,0 +1,128 @@
...
+    use crate::element::ClassifyElement;
+    use crate::link::ClassifyLink;
+    use crate::utils::test::packet_collectors::ExhaustiveCollector;
+    use crate::utils::test::packet_generators::{immediate_stream, PacketIntervalGenerator};
+    use core::time;
+    use crossbeam::crossbeam_channel;
+    use futures::future::lazy;
+
+    #[allow(dead_code)]
+    struct ClassifyEvenOddElement {
```
ddowl: This probably shouldn't be redefined every time we need to test a new link w/ classification.

route-rs-runtime/src/link/blackhole_link.rs
CollinValley: Going to drop this explanation. Will leave the original up in async_link, will pull into documentation eventually.

route-rs-runtime/src/link/blackhole_link.rs
CollinValley: Moved to a new common file in elements.

route-rs-runtime/src/link/blackhole_link.rs
```diff
@@ -0,0 +1,119 @@
...
+    use super::*;
+    use crate::element::Classifier;
+    use crate::link::ClassifyLink;
+    use crate::utils::test::packet_collectors::ExhaustiveCollector;
+    use crate::utils::test::packet_generators::{immediate_stream, PacketIntervalGenerator};
+    use core::time;
+    use crossbeam::crossbeam_channel;
+    use futures::future::lazy;
+
+    struct ClassifyEvenness {}
```
ddowl, line 41: This classifier should defined somewhere else, like near the default elements?

ddowl (approved):
LGTMMDGJ

PR #92, opened 2019-08-31
working on implementing tracing Subscriber

ddowl:
This commit introduces a simple implementation of a `tracing::Subscriber` over Sam's `trivial-identity` example. It prints new spans/events to the console as it encounters them. There are a few improvements that could be made to the subscriber, like accessing parent span context for for better more context aware info. This can be done by accessing the parent of spans/events in the trace tree. Currently, the `TrivialIdentitySubscriber` acts like an overly-complicated non-context-aware logger.


route-rs-runtime/src/element/identity_elements.rs
```diff
@@ -1,26 +1,29 @@
...
+use std::fmt::Debug;
 use std::marker::PhantomData;
+use tracing::instrument;
 
 /* IdentityElement
   This is an element that passes what it has received
 */
 #[derive(Default)]
-pub struct IdentityElement<A: Sized> {
+pub struct IdentityElement<A: Sized + Debug> {
```
ddowl, line 10: This is here because the `#[instrument]` macro treats the arguments to the function it's instrumenting as `span` metadata. Those values are required to be `Debug`.

ddowl:
Other improvements would be to instrument a `Link` `poll` fn as in [this doc example](https://docs.rs/tracing/0.1.7/tracing/span/index.html#closing-spans) showing a `span` tied to a `Future`/`Stream`. In order to do that, we'd need to uniquely identify spans by their attributes rather than assigning incremented ids for every span that's encountered.

ddowl:
For those who are interested in learning more about how `tracing` works, these docs were the most helpful:

[Overview](https://docs.rs/tracing/0.1.7/tracing/)
[Spans](https://docs.rs/tracing/0.1.7/tracing/span/index.html)
[Subscribers](https://docs.rs/tracing/0.1.7/tracing/subscriber/trait.Subscriber.html)

route-rs-runtime/src/element/identity_elements.rs
CollinValley: That's fine, we will probably use debug a lot anyways. It is a useful thing.

CollinValley (commented):
This looks involved but pretty cool, if you think it is worth pursuing we can eventually have a sprint where we  switch everything.  Though it is best done file by file I guess.

ddowl:
Yeah I think it looks promising, but I'm gonna table looking further into it to focus on the POC.

PR #93, opened 2019-09-01
WIP: Initial ICMP utility to generate error messages

eeverett6:
This is a first pass at an ICMP error message generator, and is IPv4
only for now. smoltcp does not offer a way to create all the ICMP error
types we need, so the interface is a bit complex. Some parts are
slightly hacky as well due to other smoltcp limitations.

### DIFF
--- a/route-rs-runtime/src/element/mod.rs
+++ b/route-rs-runtime/src/element/mod.rs
@@ -9,3 +9,5 @@ pub use self::transform_elements::*;
 
 mod dec_ip_hop;
 pub use self::dec_ip_hop::*;
+
+pub mod utils;
--- a/route-rs-runtime/src/element/utils/icmp.rs
+++ b/route-rs-runtime/src/element/utils/icmp.rs
@@ -1,0 +1,262 @@
+use smoltcp::phy::ChecksumCapabilities;
+use smoltcp::wire::*;
+use std::cmp;
+
+/// The size of an ICMP "header", which precedes the data field
+const ICMP_HDR_LEN: usize = 8;
+
+/// Helper to generate ICMPv4 error messages
+pub struct Icmpv4ErrorGenerator {
+    src_ip: Ipv4Address,
+    mtu: usize,
+    ttl: u8,
+    bad_source_addrs: Vec<Ipv4Address>,
+}
+
+impl Icmpv4ErrorGenerator {
+    /// Create an ICMPv4 error generator
+    ///
+    /// # Arguments
+    ///
+    /// * `src_ip` - The source IPv4 address of the generated errors
+    /// * `bad_source_addrs` - Packets from these IP addresses will generate a None instead of an
+    /// an ICMP error. Intended to contain the broadcast address of a subnet
+    pub fn new(src_ip: Ipv4Address, bad_source_addrs: Vec<Ipv4Address>) -> Icmpv4ErrorGenerator {
+        Icmpv4ErrorGenerator {
+            src_ip,
+            mtu: IPV4_MIN_MTU,
+            ttl: 64, // TODO: larger to ensure it reaches original host?
+            bad_source_addrs,
+        }
+    }
+
+    /// Create an ICMPv4 TTL exceeded error message. Return `None` if an ICMP error message should
+    /// not be generated, and the packet should be silently discarded instead
+    ///
+    /// # Arguments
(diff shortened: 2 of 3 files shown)


route-rs-runtime/src/element/mod.rs
```diff
@@ -9,3 +9,5 @@ pub use self::transform_elements::*;
 
 mod dec_ip_hop;
 pub use self::dec_ip_hop::*;
+
+pub mod utils;
```
eeverett6, line 13: This seems like the most logical place for network helpers, thoughts?

route-rs-runtime/src/element/utils/icmp.rs
```diff
@@ -0,0 +1,262 @@
...
+
+    // Performs checks based on RFC 1812 4.3.2.7 (When Not to Send ICMP Errors)
+    fn should_generate_error(&self, packet: &mut Ipv4Packet<Vec<u8>>) -> bool {
+        // Only the first fragment
+        if packet.frag_offset() != 0 {
+            return false;
+        }
+        // Avoid infinite loops, no errors from errors
+        if packet.protocol() == IpProtocol::Icmp {
+            if let Ok(icmp_packet) = Icmpv4Packet::new_checked(packet.payload_mut()) {
```
eeverett6, line 113: packet.payload() does not work with underlying type Vec[u8] but packet.payload_mut does for some reason, so we need a mutable borrow even though we don't mutate anything?

route-rs-runtime/src/element/utils/icmp.rs
```diff
@@ -0,0 +1,262 @@
...
+        error_pkt.fill_checksum();
+
+        // Generate the ICMP message
+        let mut icmp_error = Icmpv4Packet::new_unchecked(error_pkt.payload_mut());
+        icmp_error.set_msg_type(msg_type);
+        icmp_error.set_msg_code(msg_code);
+        // TODO: fill in bytes 5 - 8 with _optional data for ICMP errors that require it
+        icmp_error
+            .data_mut()
+            .clone_from_slice(&packet.clone().into_inner()[..copied_payload_len]);
```
eeverett6, line 99: Copying at IP packet at the byte level is difficult because we must consume the packet in order to get the byte-level representation of the header

route-rs-runtime/src/element/utils/icmp.rs
```diff
@@ -0,0 +1,262 @@
...
+        let mut error_pkt = Ipv4Packet::new_unchecked(buffer);
+        error_pkt_repr.emit(&mut error_pkt, &ChecksumCapabilities::ignored());
+        error_pkt.set_dscp(packet.dscp());
+        error_pkt.fill_checksum();
+
+        // Generate the ICMP message
+        let mut icmp_error = Icmpv4Packet::new_unchecked(error_pkt.payload_mut());
+        icmp_error.set_msg_type(msg_type);
+        icmp_error.set_msg_code(msg_code);
+        // TODO: fill in bytes 5 - 8 with _optional data for ICMP errors that require it
```
eeverett6, line 96: These bytes are needed for some ICMP error messages that aren't yet supported in smoltcp

route-rs-runtime/src/element/utils/icmp.rs
```diff
@@ -0,0 +1,262 @@
...
+/// Helper to generate ICMPv4 error messages
+pub struct Icmpv4ErrorGenerator {
+    src_ip: Ipv4Address,
+    mtu: usize,
+    ttl: u8,
+    bad_source_addrs: Vec<Ipv4Address>,
+}
+
+impl Icmpv4ErrorGenerator {
+    /// Create an ICMPv4 error generator
```
CollinValley, line 17: I like the documentation to start on this. Good idea since this is user facing eventually.

route-rs-runtime/src/element/utils/icmp.rs
CollinValley: Yeah it is a weird implementation detail in smoltcp. I wonder if it is just something they never thought about?

route-rs-runtime/src/element/mod.rs
CollinValley: Seems like a decent location, if it needs to be moved later, it isn't a big deal.

PR #94, opened 2019-09-02
Classify traits

scgruber:
This PR does two things:

- Restructure the trait for the element inside ClassifyLink to no longer subtrait from Element. This means we can clearly take an immutable reference to the packet and derive a class for dispatching.
- Tighten the Egressor trait bounds enough that yes we can throw away ClassifyEgressor and just use AsyncEgressor inside ClassifyLink. Yay deleting code.

### DIFF
--- a/examples/dns-interceptor/src/elements.rs
+++ b/examples/dns-interceptor/src/elements.rs
@@ -1,5 +1,5 @@
 use crate::packets::*;
-use route_rs_runtime::element::{ClassifyElement, Element};
+use route_rs_runtime::element::{Classifier, Element};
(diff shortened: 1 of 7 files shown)


ddowl (dismissed):
Nice job @scgruber !

CollinValley (approved):
LGTMMD:  This really cleans up quite a bit of code, and the interface abstraction is much easier to reason about. 
You implement something that takes in packet, provides a classification(generally in the form of an enum).  You provide a closure that does the mapping of classification to desired port number, when creating the link, and we are off to the races. Much simpler than before.

PR #95, opened 2019-09-02
Skeleton for local-dns-nat example crate

scgruber:
We're code jamming this crate, so this is a basic skeleton that we can
all work from.


CollinValley (approved):
LGTMMD

PR #96, opened 2019-09-03
Multiple input recv dedup and fairness

ErrHu:
JoinEgressor would split the from_ingressor vec into two based on
which ingressor was most recently pulled from. As it would select
the same ingressor to start pulling each time, it's not very fair.

This makes port num logic simpler and deduplicates the code and
will start recving the ingressor after the most recent pulled.


CollinValley (approved):
This is a good change, uses the power of the higher level language well to cleanup the code.  If you are feeling emboldened there other for loops that would benefit from similar treatment, though I know working with iterator chains can be a bit of a pain.
Thanks for also creating the fairness test, should catch stuff if the logic accidentally changes

PR #103, opened 2019-09-06
Docs improved, tests cleaned up, names improved.

CollinValley:
This is a big ol boy that I have been working on, but I think it does a lot of improve the runtime codebase. It uses more standard names, improves the docs in a lot of places, and provides some narrative to the code base which will help newer people understand what is happening. 

This is a big change, so I'd like feedback on the global points but I don't want to bikeshed too hard about names or small nits.  It would be better to get a baseline setup, and then accept smaller PRs, issues from here to keep the code pipeline flowing. 

Protip, the way to generate the docs the best is `cargo doc --no-deps --open --document-private-items`

### DIFF
--- a/route-rs-runtime/src/lib.rs
+++ b/route-rs-runtime/src/lib.rs
@@ -3,7 +3,26 @@ extern crate futures;
 extern crate crossbeam;
 extern crate tokio;
 
+/// Elements are the unit of transformation in route-rs. The Element is defined by a trait that requires that
+/// creators of elements implement a process function, that the underlying link will call on every packet that
+/// moves through it. You can use elements to classify, modify headers, count packets, update state, create new packets,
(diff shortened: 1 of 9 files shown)


route-rs-runtime/src/lib.rs
```diff
@@ -3,7 +3,26 @@ extern crate futures;
 extern crate crossbeam;
 extern crate tokio;
 
+/// Elements are the unit of transformation in route-rs. The Element is defined by a trait that requires that
```
scgruber, line 6: This is an annoying requirement, but Github is clipping these comments because the lines are too long. Can you make them shorter? I guess `rustfmt` doesn't mangle comment lines?

scgruber: This is an unstable option so I guess we'll have to fix our comments manually for now.

scgruber (approved):
This looks good to me, probably best to just get it in since it touches so much of the tests.

PR #104, opened 2019-09-08
graphgen: Append struct name to element identifier

scgruber:
As pipelines get bigger, it becomes really hard to remember what
"elem_1" means. Therefore, this commit appends the struct name to the
identifier so that it's easier to read later on. So that clippy doesn't
get mad about casing, the struct is lowercased in place.


scgruber:
Withdrawn in favor of #105

Issue #70, opened 2019-08-18
Support Join and Classify in graphgen


scgruber:
Right now graphgen doesn't handle any sort of joining or splitting behaviors, so it's hard to write an interesting router with it. We should support JoinElementLink and ClassifyElementLink, which requires a bunch of thought about how to autogenerate the dispatching logic.

PR #105, opened 2019-09-09
Graphgen Join & Classify

### LINKED ISSUE #69: Build integration tests for graphgen
Graphgen creates pipelines from a graph file. Right now, we don't have any tests that this continues to work, and we only have one example crate to test against. We should have an extensible testing framework for these and more examples.

scgruber:
Supports Join & Classify in graphgen and adds a test for `examples/dns-interceptor` now that we can generate all of it.

Fixes #69 and #70


CollinValley (approved):
Looks good to me, did not do a super close review, I trust you with your own subsystem. But everything seems normal, good stuff.

PR #106, opened 2019-09-10
Improve flakiness of join_element_fairness_test

ErrHu:
The fairness test should cycle through both generators one after another
but the order is not guaranteed. This modifies the test to see if all
packets from the 1 generator (4 packets) are received within the 1st 10
joins.


ddowl (approved):
This seems fine to me. It's just ensuring that not all the 0s are getting served before 1s. If we need to be more strict about the definition of fairness in the future, we can update the test then.

PR #138, opened 2019-09-13
cargo: pin crate dependencies

### LINKED ISSUE #91: Discuss Pinning all our packages
We need to fully debate the best way to denote the versions of our dependencies in our cargo.toml files. Ideally we would set a specific version, maybe we could use more of a meta-version if they are available 'latest-stable' or whatever.

ddowl:
fixes #91


CollinValley (approved):
LGTMMD
There are some other things we can do, like specify a version range, but keeping it exact is easier, especially in the early stages.  I like the exact versions so we are all on the same page. 
For the curious: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html

PR #144, opened 2019-09-15
Element: change 'process' to return an Option<Self::Output>

### LINKED ISSUE #102: Allow elements to drop packets
Right now, elements have no way to drop a packet, since they are forced to return a packet to the link due to the way their process function is implemented. We should change that, so that process returns the Option<packet>.  Then links will unwrap the Option and forward the packet if it exists, if they get a None they will re-fetch and retry.

ddowl:
This commit changes the `process` function of `Element` and `AsyncElement` to return an `Option<Self::Output>` rather than `Self::Output`, so they can support packet drops by returning `None`.

`Link`s will now continue polling their input Streams until their either the Stream is exhausted, or its `Element` returns `Some(packet)`.

A `DropElement` is implemented in a similar fashion to `IdentityElement`, where it drops every packet that it receives, rather than passes it unchanged.

resolves #102

### DIFF
--- a/route-rs-runtime/src/element/drop_element.rs
+++ b/route-rs-runtime/src/element/drop_element.rs
@@ -1,0 +1,36 @@
+use crate::element::{AsyncElement, Element};
+use std::marker::PhantomData;
+
+/* DropElement
+  This element drops every packet that it receives
+*/
+#[derive(Default)]
+pub struct DropElement<A: Sized> {
+    phantom: PhantomData<A>,
+}
+
+impl<A> DropElement<A> {
+    pub fn new() -> DropElement<A> {
+        DropElement {
+            phantom: PhantomData,
+        }
+    }
+}
+
+impl<A> Element for DropElement<A> {
+    type Input = A;
+    type Output = A;
+
+    fn process(&mut self, _packet: Self::Input) -> Option<Self::Output> {
+        None
+    }
+}
+
+impl<A> AsyncElement for DropElement<A> {
+    type Input = A;
+    type Output = A;
+
+    fn process(&mut self, _packet: Self::Input) -> Option<Self::Output> {
+        None
+    }
+}
--- a/route-rs-runtime/src/link/async_link.rs
+++ b/route-rs-runtime/src/link/async_link.rs
@@ -95,21 +95,28 @@ impl<E: AsyncElement> Future for AsyncIngressor<E> {
 
     /// Implement Poll for Future for AsyncIngressor
     ///
-    /// Note that this function works a bit different, it continues to process
+    /// This function continues to process
     /// packets off it's input queue until it reaches a point where it can not
-    /// make forward progress. There are three cases:
+    /// make forward progress. There are several cases:
     /// ###
     /// #1 The to_egressor queue is full, we notify the Egressor that we need
-    /// awaking when there is work to do, and go to sleep.
+    /// awaking when there is work to do, and go to sleep by returning `Async::NotReady`.
     ///
     /// #2 The input_stream returns a NotReady, we sleep, with the assumption
     /// that whomever produced the NotReady will awaken the task in the Future.
     ///
     /// #3 We get a Ready(None), in which case we push a None onto the to_Egressor
     /// queue and then return Ready(()), which means we enter tear-down, since there
-    /// is no futher work to complete.
-    /// ###
-    /// By Sleep, we mean we return a NotReady to the runtime which will sleep the task.
+    /// is no further work to complete.
+    ///
+    /// #4 If our upstream `PacketStream` has a packet for us, we pass it to our `element`
+    /// for `process`ing. Most of the time, it will yield a `Some(output_packet)` that has
+    /// been transformed in some way. We pass that on to our egress channel and notify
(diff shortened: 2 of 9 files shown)


ddowl:
What becomes of `BlackHoleLink` if all elements support packet drops?

@CollinValley

CollinValley:
> What becomes of `BlackHoleLink` if all elements support packet drops?
> 
> @CollinValley

Blackhole link is a way to cap a classification/stream. It still has uses especially when you have unimplemented parts of the router you would like to leave a connection for in the future. Consider  Ipv6 support, or a weird transport layer protocol. Classify doesn't drop packets.  I think supporting both is a good idea, since each is more ergonomic in various situations.

route-rs-runtime/src/link/async_link.rs
```diff
@@ -120,13 +120,15 @@ impl<E: AsyncElement> Future for AsyncIngressor<E> {
...
                 None => return Ok(Async::Ready(())),
-                Some(input_packet) => {
-                    let output_packet: E::Output = self.element.process(input_packet);
-                    self.to_egressor
-                        .try_send(Some(output_packet))
-                        .expect("AsyncIngressor::Poll: try_send to_egressor shouldn't fail");
-                    unpark_and_notify(&self.task_park);
-                }
+                Some(input_packet) => match self.element.process(input_packet) {
+                    None => return Ok(Async::Ready(())),
```
CollinValley: This is incorrect behavior.  Consider that Ok(Async::Ready()) should be returned when the link is ready to be torn down by Tokio.  So in the case we process a packet and the element returns it as invalid, we should do nothing and go around the loop again, since there still may be more packets to pull and process.

route-rs-runtime/src/link/sync_link.rs
```diff
@@ -44,8 +44,8 @@ impl<E: Element> Stream for SyncLink<E> {
         match input_packet_option {
             None => Ok(Async::Ready(None)),
             Some(input_packet) => {
-                let output_packet: E::Output = self.element.process(input_packet);
-                Ok(Async::Ready(Some(output_packet)))
+                let output_packet_option: Option<E::Output> = self.element.process(input_packet);
```
CollinValley: Same as a above, we don't want to return the None since that is a teardown signal. See line 29 docs.  Instead we should silently drop the None, and pull another packet for processing, until we exhaust the upstream, or have a valid packet to return.

CollinValley: You should also update the documentation above here to note that in the Ready(Some) case, the packet is processed by the element, which may or may not yield a valid return.

route-rs-runtime/src/element/drop_element.rs
```diff
@@ -0,0 +1,36 @@
+use crate::element::{AsyncElement, Element};
```
CollinValley, line 1: For good test coverage, you should also implement an element that only drops some packets. This way we can ensure the case I mentioned above is actually handled.   Maybe a random drop element since that would be useful in real routers to simulate a bad network connection.

CollinValley (commented):
The concept is good, I thought about Option vs Result, I think Option makes sense here because the link shouldn't concern itself with an Err value anyways. Links are about what-to-where rather than why in the router. 
See my comments for the required changes.

route-rs-runtime/src/link/async_link.rs
ddowl: Does that mean recursively invoking `poll` or deferring with `Ok(Async::NotReady)`?

ddowl: > we should do nothing and go around the loop again

Pretty sure this translates to
```
park_and_notify(&self.task_park);
return Ok(Async::NotReady);
```

ddowl: Damn dawg this poll has a `loop` in it my b.

route-rs-runtime/src/element/drop_element.rs
ddowl: Thats a good idea! And probably not hard to test if we can set its PRG seed.

ddowl: Created #145 to track that.

route-rs-runtime/src/link/sync_link.rs
```diff
@@ -20,32 +20,37 @@ impl<E: Element> Stream for SyncLink<E> {
...
-    "try_ready!" macro, which will automatically return "Ok(Async::NotReady)" if the input stream
-    gives us NotReady.
-
-    Err: is also handled by the "try_ready!" macro.
-    */
+    /// Intro to `Stream`s:
+    /// 4 cases: `Async::Ready(Some)`, `Async::Ready(None)`, `Async::NotReady`, `Err`
+    ///
+    /// `Async::Ready(Some)`: We have a packet ready to process from the upstream element.
+    /// It's passed to our core's process function for... processing
```
CollinValley, line 27: Nit:  switch core to element in the docs.

CollinValley (approved):
Just a Nit, optional fix
Otherwise LGTMMD!

PR #146, opened 2019-09-26
Link: add Link trait and impl Link for BlackHoleLink

ddowl:
This is the first stab at setting up a common, composable interface for
Links, and a POC implementation with BlackHoleLink (the easiest to
impl).

### DIFF
--- a/route-rs-runtime/src/link/blackhole_link.rs
+++ b/route-rs-runtime/src/link/blackhole_link.rs
@@ -4,48 +4,53 @@ use futures::{Async, Future, Poll};
 /// Link that drops all packets ingressed.
 #[derive(Default)]
 pub struct BlackHoleLink<Packet: Sized> {
-    ingress_streams: Option<Vec<PacketStream<Packet>>>,
+    hole: Option<BlackHole<Packet>>,
 }
 
 impl<Packet> BlackHoleLink<Packet> {
     pub fn new() -> Self {
-        BlackHoleLink {
-            ingress_streams: None,
-        }
-    }
-}
-
-impl<Packet: Sized> Future for BlackHoleLink<Packet> {
-    type Item = ();
-    type Error = ();
-
-    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
-        match self.ingress_streams.as_mut() {
-            None => panic!("Cannot poll! Missing ingress streams"),
-            Some(ingress_streams) => loop {
-                for ingress_stream in ingress_streams.iter_mut() {
-                    if try_ready!(ingress_stream.poll()).is_none() {
-                        return Ok(Async::Ready(()));
-                    }
-                }
-            },
-        }
+        BlackHoleLink { hole: None }
     }
 }
 
 impl<Packet: Sized + 'static> Link<Packet, ()> for BlackHoleLink<Packet> {
     fn ingressors(&self, ingress_streams: Vec<PacketStream<Packet>>) -> Self {
         BlackHoleLink {
-            ingress_streams: Some(ingress_streams),
+            hole: Some(BlackHole::new(ingress_streams)),
         }
     }
 
     fn build_link(self) -> (Vec<TokioRunnable>, Vec<PacketStream<()>>) {
-        if self.ingress_streams.is_none() {
+        if self.hole.is_none() {
             panic!("Cannot build link! Missing ingress streams")
         }
 
-        (vec![Box::new(self)], vec![])
+        (vec![Box::new(self.hole.unwrap())], vec![])
+    }
+}
+
+pub struct BlackHole<Packet> {
+    ingress_streams: Vec<PacketStream<Packet>>,
+}
+
+impl<Packet: Sized> BlackHole<Packet> {
+    fn new(ingress_streams: Vec<PacketStream<Packet>>) -> Self {
+        BlackHole { ingress_streams }
+    }
+}
+
+impl<Packet: Sized> Future for BlackHole<Packet> {
+    type Item = ();
+    type Error = ();
+
+    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+        loop {
+            for ingress_stream in self.ingress_streams.iter_mut() {
+                if try_ready!(ingress_stream.poll()).is_none() {
+                    return Ok(Async::Ready(()));
+                }
+            }
+        }
     }
 }
 
@@ -141,18 +146,20 @@ mod tests {
 
         let drain0 = link0.ingressor;
 
-        let link1 = BlackHoleLink::new().ingressors(vec![Box::new(link0.egressors.pop().unwrap())]);
+        let (mut black_hole_runnables, _) = BlackHoleLink::new()
+            .ingressors(vec![Box::new(link0.egressors.pop().unwrap())])
+            .build_link();
 
         let (s0, link0_port0_collector_output) = crossbeam_channel::unbounded();
         let link0_port0_collector =
             ExhaustiveCollector::new(0, Box::new(link0.egressors.pop().unwrap()), s0);
 
-        tokio::run(lazy(|| {
-            tokio::spawn(drain0);
-            tokio::spawn(link0_port0_collector);
-            tokio::spawn(link1);
-            Ok(())
-        }));
+        let mut runnables: Vec<TokioRunnable> = Vec::new();
+        runnables.push(Box::new(drain0));
+        runnables.push(Box::new(link0_port0_collector));
+        runnables.append(&mut black_hole_runnables);
+
+        run_tokio(runnables);
 
         let elem0_port0_output: Vec<_> = link0_port0_collector_output.iter().collect();
         assert_eq!(elem0_port0_output, vec![0, 2, 420, 4, 6, 8]);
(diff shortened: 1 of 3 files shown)


route-rs-runtime/src/link/mod.rs
```diff
@@ -33,7 +33,14 @@ mod blackhole_link;
 pub use self::blackhole_link::*;
 
 /// The type that all links take as input, or provide as output. This allows links to be combinatorial
-pub type PacketStream<Input> = Box<dyn Stream<Item = Input, Error = ()> + Send>;
+pub type PacketStream<Input> = Box<dyn Stream<Item = Input, Error = ()> + 'static + Send>;
```
ddowl, line 36: Added `'static` since these `Streams` should always last the lifetime of the router.

route-rs-runtime/src/link/mod.rs
```diff
@@ -33,7 +33,14 @@ mod blackhole_link;
...
+pub type TokioRunnable = Box<dyn Future<Item = (), Error = ()> + 'static + Send>;
```
ddowl, line 38: I'd like to not have the `Future` be boxed, but was running into the `trait objects without an explicit 'dyn' are deprecated` stuff.

route-rs-runtime/src/link/mod.rs
```diff
@@ -33,7 +33,14 @@ mod blackhole_link;
...
+
+pub trait Link<A, B> {
+    fn new(ingress_streams: Vec<PacketStream<A>>) -> Self;
+    fn runnables(self) -> Vec<TokioRunnable>;
```
ddowl: Since `runnables` should be the last thing that's called on a `Link` I'm making it consume `self`. This is needed if we want to pass ownership of the object (which is also the `TokioRunnable`) to tokio.

route-rs-runtime/src/link/blackhole_link.rs
```diff
@@ -55,14 +63,23 @@ mod tests {
         }
     }
 
+    fn run_tokio(runnables: Vec<TokioRunnable>) {
```
ddowl, line 85: This shit just works! :tada:

route-rs-runtime/src/link/blackhole_link.rs
```diff
@@ -18,15 +12,29 @@ impl<Packet: Sized> Future for BlackHoleLink<Packet> {
...
     }
 }
 
+impl<Packet: Sized + 'static> Link<Packet, ()> for BlackHoleLink<Packet> {
+    fn new(ingress_streams: Vec<PacketStream<Packet>>) -> Self {
+        BlackHoleLink { ingress_streams }
+    }
+
+    fn runnables(self) -> Vec<TokioRunnable> {
+        vec![Box::new(self)]
```
ddowl: I'd just like to pass out `self`, but running into the `doesn't have a size known at compile-time` stuff.

route-rs-runtime/src/link/blackhole_link.rs
```diff
@@ -18,15 +12,29 @@ impl<Packet: Sized> Future for BlackHoleLink<Packet> {
 
     fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
         loop {
-            let input_packet_option: Option<Packet> = try_ready!(self.input_stream.poll());
-            match input_packet_option {
-                None => return Ok(Async::Ready(())),
-                Some(_) => {}
-            };
+            for ingress_stream in self.ingress_streams.iter_mut() {
```
ddowl, line 48: Tried using `self.ingress_streams.iter_mut().cycle()` as my iterator, but requires its elements to be `Clone` for some reason. I don't think it's too bad to just keep the loop like this though.

ddowl:
Added some docstrings to the `Link` trait and associated types.

route-rs-runtime/src/link/mod.rs
CollinValley: Is this required that they be static? I am fine with it but usually static is a sign that you could have a better lifetime bound.

route-rs-runtime/src/link/blackhole_link.rs
CollinValley: There is probably some good reason. I think this loop is okay.

route-rs-runtime/src/link/blackhole_link.rs
CollinValley: Makes sense size is not known cause what we are in this case is a Vec<PacketStream<Packet>>.  Boxing it up is fine. I mean, kind of makes sense all our static lifetimed shit lives on the heap anyways.

route-rs-runtime/src/link/blackhole_link.rs
CollinValley: :+100:

route-rs-runtime/src/link/mod.rs
CollinValley: We can always try to remove static later though, it doesn't seem like an issue.

route-rs-runtime/src/link/mod.rs
CollinValley: Yeah, I figured this would happen.  Good news though is that the function call stack for it will look something like 
poll()..poll()..poll()..poll(). Before, the last 3 polls where already dyn since they were streams.  Now they are all dyn. Doesn't seem like a huge cost to me for the flexibility we gain. I assume the CPU hardware will do a good job prefetching the fn instructions, especially when you do it quickly on repeat over and over.

CollinValley: Ideally we wouldn't need Dyn's a guess? One thing to consider though is that if we do manage to monomorph our functions, the size of the router may go up quite a bit.

route-rs-runtime/src/link/mod.rs
CollinValley: I think we are discussing this in the discord. While this is pretty good. Moving over to a builder syntax that build consumes self may be more ergonomic.  I assume that egressors would also need to consume self, since it will move the objects out as well.

route-rs-runtime/src/link/mod.rs
```diff
@@ -32,8 +32,26 @@ pub use self::tee_link::*;
...
+/// All Links communicate through streams of packets. This allows them to be composable.
+pub type PacketStream<Input> = Box<dyn Stream<Item = Input, Error = ()> + 'static + Send>;
+/// Some Links may need to be driven by Tokio. This represents a handle to something Tokio can run.
+pub type TokioRunnable = Box<dyn Future<Item = (), Error = ()> + 'static + Send>;
+
+/// All *Link's should implement Link in order to be composable with each other.
+///
+/// The two type parameters, A and B, refer to the input and output types of the Link.
+/// You can tell because the ingress/egress streams are of type `PacketStream<A>`/`PacketStream<B>` respectively.
+pub trait Link<A, B> {
```
CollinValley: Lets change out A and B for Input and Output.

ddowl:
Uploaded another try using the builder pattern. I also commented out the graphgen tests while the API is in flux. `Link` instances look a lot cleaner now! There are still a few things I have questions about:

1. So far, we've been implementing `Future` and `Stream` over several Link structs, sans those that need queues, and passing them into `tokio::run`. This creates a weird situation when building s.t. you need to check for valid `ingress_streams` in `poll`, since it may or may not have been populated by `Link::ingressors`. This is an impossible state though, as `build_link` should panic if the all the required inputs aren't available at that time. Therefore, I'd like `Link`s to own separate `TokioRunnable` objects that implement `Future` themselves (if needed), so we can simplify the code. I'll make a separate commit/PR for that.

2. I learned that if a struct impls a  `fn new() -> Self`, clippy will suggest that it derives `Default`. At first, this didn't seem to fit the builder pattern here, but then I realized how cool it would be if the `Link` trait had a default implementation of `new() -> Self` that returned each object set to its initial state (likely a bunch of `None`s). I couldn't quite figure out how to get that working, or if that's possible.

route-rs-runtime/src/link/mod.rs
CollinValley: Resolving this as fine 🔥 👌

route-rs-runtime/src/link/blackhole_link.rs
```diff
@@ -1,28 +1,55 @@
...
+    fn build_link(self) -> (Vec<TokioRunnable>, Vec<PacketStream<()>>) {
+        if self.hole.is_none() {
+            panic!("Cannot build link! Missing ingress streams")
+        }
+
+        (vec![Box::new(self.hole.unwrap())], vec![])
+    }
+}
+
+pub struct BlackHole<Packet> {
```
CollinValley, line 32: Potential Nit: Should this be called BlackHoleIngressor? It is awkward naming here, but in all the other links the thing that implements the `TokioRunnable` should be the Ingressor, and the thing that Impls `Stream` would be the egressor?

CollinValley: Nit followup: Implements is awkward phrasing, the thing that is a TokioRunnable object IE, the struct that implements Future, is the ingressor in all other cases.  I think we can leave this as is for now, and revisit later once the other implementations shake out. The names my clarify. I think leaving it blackhole is fine in either case. Just wanted to make the suggestion.

CollinValley (approved):
LGTMMD
I am also opening an issue to add multiple input stream test to BlackHole, since we now support that dope functionality.

route-rs-runtime/src/link/blackhole_link.rs
ddowl: Yep, we should revisit namings when more of the Links have been migrated. I went back and forth between a few ideas in the next commit(s).

PR #162, opened 2019-10-03
SyncLink impls Links, minor changes to Link API and BlackHole

ddowl:
Commit messages are pretty descriptive, but this PR takes the next leap to have `SyncLink` impl `Link`, or rather a `SyncLinkBuilder` impl `LinkBuilder`. It's up for discussion, but I think having an explicit `*Builder` makes sense since the purpose of these objects are now to keep track of enough inputs to construct some `Link`, which is now defined as the product type of the necessary `TokioRunnables` and `PacketStream<Output>` (i.e. egressors) that need to be driven and exposed, respectively. The name `Link` in particular is confusing because of its type declaration `Link<Output>`, which isn't totally clear. I've also disabled `graphgen` tests.

### DIFF
--- a/route-rs-runtime/src/link/blackhole_link.rs
+++ b/route-rs-runtime/src/link/blackhole_link.rs
@@ -14,7 +14,7 @@ impl<Packet> BlackHoleLink<Packet> {
 }
 
 impl<Packet: Sized + 'static> Link<Packet, ()> for BlackHoleLink<Packet> {
-    fn ingressors(&self, ingress_streams: Vec<PacketStream<Packet>>) -> Self {
+    fn ingressors(self, ingress_streams: Vec<PacketStream<Packet>>) -> Self {
         BlackHoleLink {
             hole: Some(BlackHole::new(ingress_streams)),
         }
@@ -22,20 +22,20 @@ impl<Packet: Sized + 'static> Link<Packet, ()> for BlackHoleLink<Packet> {
 
     fn build_link(self) -> (Vec<TokioRunnable>, Vec<PacketStream<()>>) {
         if self.hole.is_none() {
-            panic!("Cannot build link! Missing ingress streams")
+            panic!("Cannot build link! Missing input streams");
+        } else {
+            (vec![Box::new(self.hole.unwrap())], vec![])
         }
-
-        (vec![Box::new(self.hole.unwrap())], vec![])
     }
 }
 
-pub struct BlackHole<Packet> {
-    ingress_streams: Vec<PacketStream<Packet>>,
+struct BlackHole<Packet> {
+    in_streams: Vec<PacketStream<Packet>>,
 }
 
 impl<Packet: Sized> BlackHole<Packet> {
-    fn new(ingress_streams: Vec<PacketStream<Packet>>) -> Self {
-        BlackHole { ingress_streams }
+    fn new(in_streams: Vec<PacketStream<Packet>>) -> Self {
+        BlackHole { in_streams }
     }
 }
 
@@ -45,8 +45,8 @@ impl<Packet: Sized> Future for BlackHole<Packet> {
 
     fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
         loop {
-            for ingress_stream in self.ingress_streams.iter_mut() {
-                if try_ready!(ingress_stream.poll()).is_none() {
+            for in_stream in self.in_streams.iter_mut() {
+                if try_ready!(in_stream.poll()).is_none() {
                     return Ok(Async::Ready(()));
                 }
             }
--- a/route-rs-runtime/src/link/blackhole_link.rs
+++ b/route-rs-runtime/src/link/blackhole_link.rs
@@ -4,27 +4,30 @@ use futures::{Async, Future, Poll};
 /// Link that drops all packets ingressed.
 #[derive(Default)]
 pub struct BlackHoleLink<Packet: Sized> {
-    hole: Option<BlackHole<Packet>>,
+    in_streams: Option<Vec<PacketStream<Packet>>>,
 }
 
 impl<Packet> BlackHoleLink<Packet> {
     pub fn new() -> Self {
-        BlackHoleLink { hole: None }
+        BlackHoleLink { in_streams: None }
     }
 }
 
 impl<Packet: Sized + 'static> Link<Packet, ()> for BlackHoleLink<Packet> {
     fn ingressors(self, ingress_streams: Vec<PacketStream<Packet>>) -> Self {
         BlackHoleLink {
-            hole: Some(BlackHole::new(ingress_streams)),
+            in_streams: Some(ingress_streams),
         }
     }
 
     fn build_link(self) -> (Vec<TokioRunnable>, Vec<PacketStream<()>>) {
-        if self.hole.is_none() {
+        if self.in_streams.is_none() {
             panic!("Cannot build link! Missing input streams");
         } else {
-            (vec![Box::new(self.hole.unwrap())], vec![])
+            (
+                vec![Box::new(BlackHole::new(self.in_streams.unwrap()))],
+                vec![],
+            )
         }
     }
 }
--- a/route-rs-runtime/src/link/blackhole_link.rs
+++ b/route-rs-runtime/src/link/blackhole_link.rs
@@ -1,26 +1,26 @@
-use crate::link::{Link, PacketStream, TokioRunnable};
+use crate::link::{Link, LinkBuilder, PacketStream};
 use futures::{Async, Future, Poll};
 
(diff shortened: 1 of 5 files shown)


route-rs-runtime/src/link/mod.rs
```diff
@@ -31,25 +31,35 @@ pub use self::tee_link::*;
...
 mod blackhole_link;
 pub use self::blackhole_link::*;
+use crate::element::Element;
 
 /// All Links communicate through streams of packets. This allows them to be composable.
 pub type PacketStream<Input> = Box<dyn Stream<Item = Input, Error = ()> + 'static + Send>;
 /// Some Links may need to be driven by Tokio. This represents a handle to something Tokio can run.
 pub type TokioRunnable = Box<dyn Future<Item = (), Error = ()> + 'static + Send>;
+/// LinkBuilders build this.
+pub type Link<Output> = (Vec<TokioRunnable>, Vec<PacketStream<Output>>);
```
ddowl, line 41: Better name pls?

route-rs-runtime/src/link/sync_link.rs
```diff
@@ -1,22 +1,73 @@
...
+/// may only have one ingress and egress stream since it lacks some kind of queue
+/// storage. In the future we might decide to restrict the interface for this link
+/// for clearer intent.
+impl<E: Element + Send + 'static> LinkBuilder<E::Input, E::Output> for SyncLinkBuilder<E> {
+    fn ingressors(self, in_streams: Vec<PacketStream<E::Input>>) -> Self {
+        assert_eq!(in_streams.len(), 1, "SyncLink may only take 1 input stream");
+
+        SyncLinkBuilder {
+            in_streams: Some(in_streams),
+            element: self.element,
```
ddowl, line 37: I think this pattern of builder methods passing all of `self`'s state except for the input is going to become common. I'm not sure if anything can be done to abstract that idea, but we should wait until every `Link` is migrated before trying.

route-rs-runtime/src/link/sync_link.rs
```diff
@@ -1,22 +1,73 @@
...
+        }
+    }
+}
+
+/// Although `Link` allows an arbitrary number of ingressors and egressors, `SyncLink`
+/// may only have one ingress and egress stream since it lacks some kind of queue
+/// storage. In the future we might decide to restrict the interface for this link
+/// for clearer intent.
+impl<E: Element + Send + 'static> LinkBuilder<E::Input, E::Output> for SyncLinkBuilder<E> {
+    fn ingressors(self, in_streams: Vec<PacketStream<E::Input>>) -> Self {
```
ddowl: Do we want to enforce that each builder method should be called at most once? For example, we could assert here that `self.in_streams` must be `None`, otherwise panic with a helpful message.

route-rs-runtime/src/link/sync_link.rs
```diff
@@ -1,22 +1,73 @@
...
+    fn element(self, element: E) -> Self {
+        SyncLinkBuilder {
+            in_streams: self.in_streams,
+            element: Some(element),
+        }
+    }
+}
+
+/// The single egressor of SyncLink
+struct SyncProcessor<E: Element> {
```
ddowl, line 63: The idea behind "Processor" is that this is the thing that performs the transformation based on the `Link`'s `Element::process`, rather than just a handle for outgoing packets. It's a little strange since this entity is semantically different in different contexts: inside `SyncLink` it's a packet processor, but outside `SyncLink` it's an upstream `PacketStream` handle.

route-rs-runtime/src/link/blackhole_link.rs
```diff
@@ -1,41 +1,44 @@
...
 }
 
-impl<Packet: Sized + 'static> Link<Packet, ()> for BlackHoleLink<Packet> {
-    fn ingressors(&self, ingress_streams: Vec<PacketStream<Packet>>) -> Self {
-        BlackHoleLink {
-            hole: Some(BlackHole::new(ingress_streams)),
+impl<Packet: Sized + 'static> LinkBuilder<Packet, ()> for BlackHoleLinkBuilder<Packet> {
+    fn ingressors(self, ingress_streams: Vec<PacketStream<Packet>>) -> Self {
+        BlackHoleLinkBuilder {
+            in_streams: Some(ingress_streams),
```
ddowl, line 19: With multiple builder functions, the pattern of "accumulate all of your state, then initialize in `build_link`" became clearer, so `BlackHoleLink` is updated to reflect that.

route-rs-runtime/src/link/sync_link.rs
CollinValley: I think the question is: Does running the function twice cause issues other than the obvious  'The values change' answer.  If yes, then we should prevent double running.

ddowl: I mean, no it shouldn't, but we should be able to quickly detect nonsensical builders like:

```
SyncLinkBuilder
  .ingressors(<some vec>)
  .ingressors(<wait... another vec??>)
  .element(e)
  .build_link();
```

I guess running the function twice causes the issue of "being confusing to the reader". Functionally the last builder value is used.

CollinValley: Hmm, if you want to make a double call panic, I am fine with that.

route-rs-runtime/src/link/mod.rs
CollinValley: I mean, it isn't the worst name right now.  LinkBuilder's build Links, that is pretty clear? Still open to ideas though.

route-rs-runtime/src/link/sync_link.rs
CollinValley: I like the processor name, it tells us what it does in the code. And to the user, they never see the name. They only really interact with it through the TokioRunnables and Egressors. Seems like a good name to me.

route-rs-runtime/src/link/sync_link.rs
```diff
@@ -42,13 +93,15 @@ impl<E: Element> Stream for SyncLink<E> {
...
     fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
         loop {
-            let input_packet_option: Option<E::Input> = try_ready!(self.input_stream.poll());
-            match input_packet_option {
-                None => return Ok(Async::Ready(None)),
-                Some(input_packet) => {
-                    // if `element.process` returns None, do nothing, loop around and try polling again.
-                    if let Some(output_packet) = self.element.process(input_packet) {
-                        return Ok(Async::Ready(Some(output_packet)));
+            for ingress_stream in self.in_streams.iter_mut() {
```
CollinValley: Issue here: Since we are servicing multiple inputs, we need to fairly pull from them. If you look at join link, you want similar functionality to that.  The trick is usually to just save the index of the last thing you pulled a packet form internally, then restart the for loop at that index on the next poll.

CollinValley (changes requested):
Overall, the actual changes look really good. With the for loop functionality fixed I think this should be ready for merge.

route-rs-runtime/src/link/sync_link.rs
CollinValley: You probably want something along the lines of:
```
    fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
        //rotate_slice exists in 1.22 nightly experimental
        let rotated_iter = self
            .from_ingressors
            .iter()
            .enumerate()
            .cycle()
            .skip(self.next_pull_ingressor)
            .take(self.from_ingressors.len());
        for (port, from_ingressor) in rotated_iter {
            match from_ingressor.try_recv() {
                Ok(Some(packet)) => {
                    unpark_and_notify(&self.task_parks[port]);
                    self.next_pull_ingressor = port + 1;
                    return Ok(Async::Ready(Some(packet)));
                }
                Ok(None) => {
                    //Got a none from a consumer that has shutdown
                    self.ingressors_alive -= 1;
                    if self.ingressors_alive == 0 {
                        return Ok(Async::Ready(None));
                    }
                }
                Err(_) => {
                    //On an error go to next channel.
                }
            }
        }
```

ddowl: Offline, we spent a meetup figuring out that fairly polling many ingress streams requires some kind of intermediary channel and task_park like `JoinLink` has. This is the reason why we decided `SyncLink`s should be 1 to 1 rather than n to 1.

Since `AsyncLink` will need to fairly poll its ingress streams, fair joining functionality can be implemented there.

ddowl:
Before merging this, I'd like to make a few changes to `SyncLink` to simplify polling logic since we are guaranteed to only have 1 ingress stream.

CollinValley (approved):
LGTMMD

PR #164, opened 2019-10-08
Packets working branch

CollinValley:
This is a huge commit, everyone please go through it and point out anything that is hacky, could use improvement. Particularly, I want to weed out things that likely don't do what you expect. I want this library to not easily panic, and have a what you see is what you get feel to it.  This is a big ol' boy, and I'm going to want to merge it not completely perfect, but I think it is in a functional enough state now that it could be used to begin building elements.  Once we have a list of fixes etc, I'll pare down and clean up the commit messages.


route-rs-runtime/src/lib.rs
```diff
@@ -23,6 +23,8 @@ pub mod link;
 /// program takes the user provided graph, and generates a pipeline. It connects input and output channels to the pipe, so that it can be
 /// hooked up to packet sources and sinks, and drops the pipeline into the runtime. Generally can be abstracted to the concept, router.
 pub mod pipeline;
+pub mod packet;
```
moosingin3space: I think this could work better as a standalone crate -- that way I could use it in `afpacket` tests, for instance.

route-rs-runtime/src/packet/ipv6.rs
```diff
@@ -0,0 +1,336 @@
...
+pub struct Ipv6Packet<'packet> {
+    pub data: PacketData<'packet>,
+    // There may be various "Extension Headers", so we should figure out the actual offset and store it here for
+    // easy access in the helper functions.
+    pub packet_offset: usize,
+    pub payload_offset: usize,
+}
+
+impl<'packet> Ipv6Packet<'packet> {
+    fn new(packet: PacketData) -> Result<Ipv6Packet, &'static str> {
```
CollinValley: Forgot to switch this so the new function takes the variable packet_offset length. Will get that tomorrow.

CollinValley: See ipv4::new() for the idea.

route-rs-runtime/src/packet/ethernet.rs
```diff
@@ -0,0 +1,113 @@
+use std::borrow::Cow;
+use crate::packet::*;
+
+pub struct EthernetFrame<'packet> {
+    pub data: PacketData<'packet>,
+}
+
+impl<'frame> EthernetFrame<'frame> {
+    pub fn new(frame: PacketData) -> Result<EthernetFrame, &'static str> {
```
moosingin3space: Let's think about a legit `Error` type. I'm not too sure what the best way to do `Error`s is since `failure` became superseded by the stdlib though.

route-rs-runtime/src/packet/ethernet.rs
```diff
@@ -0,0 +1,113 @@
...
+        self.data.truncate(12); //Cut off and drop the entire payload
+        let payload_len_bytes = payload_len.to_be_bytes();
+        self.data.reserve_exact(payload_len as usize + 2); //Reserve room for payload and len field.
+        self.data.extend(payload_len_bytes.iter());
+        self.data.extend(payload);
+    }
+
+    // This is super uggo but I don't care right now, I'm sure this is a cleaner way to do it.
+    pub fn set_dest_mac(&mut self, mac: MacAddr) {
+        self.data[0] = mac.b0;
```
moosingin3space: If `MacAddr` stored an array internally, you could use [this](https://doc.rust-lang.org/std/primitive.slice.html#method.copy_from_slice).

route-rs-runtime/src/packet/ethernet.rs
```diff
@@ -0,0 +1,113 @@
...
+        self.data[6] = mac.b0;
+        self.data[7] = mac.b1;
+        self.data[8] = mac.b2;
+        self.data[9] = mac.b3;
+        self.data[10] = mac.b4;
+        self.data[11] = mac.b5;
+    }
+}
+
+impl<'packet> From<TcpSegment<'packet>> for Result<EthernetFrame<'packet>, &'static str> {
```
moosingin3space: Any reason for `From` instead of `TryFrom` here?

route-rs-runtime/src/packet/ipv4.rs
```diff
@@ -0,0 +1,72 @@
...
+
+pub struct Ipv4Packet<'packet> {
+    pub data: PacketData<'packet>,
+}
+
+impl<'packet> Ipv4Packet<'packet> {
+    fn new(packet: PacketData) -> Result<Ipv4Packet, &'static str> {
+        //Header of Ethernet Frame: 14 bytes
+        //Header of IPv4 Frame: 20 bytes
+        if packet.len() < 14 + 20 {
```
moosingin3space: Ideally a packet type doesn't have to worry about what its containing packet type is. This is worth discussion.

route-rs-runtime/src/packet/packet.rs
```diff
@@ -0,0 +1,23 @@
+//Let's use this area for now to declare common structs, constants, and common helper functions.
+#[derive(Eq, Clone, Copy, Hash, PartialEq)]
+pub struct MacAddr {
```
moosingin3space: See above comments about `copy_from_slice`.

route-rs-runtime/src/packet/udp.rs
```diff
@@ -0,0 +1,70 @@
+use crate::packet::*;
+
+pub struct UdpSegment<'packet> {
```
moosingin3space: Not sure "segment" is the semantically-correct name for this -- `UdpDatagram` maybe?

moosingin3space (changes requested):
I like this API. That said, it'd be much easier to read with constants, and I'd like to have a discussion about ways we can potentially make each inner packet independent of the outer packet storing it.

route-rs-runtime/src/packet/ethernet.rs
CollinValley: Yeah, I would like to be able to return formatted strings or something.

route-rs-runtime/src/packet/udp.rs
CollinValley: In TCP they are called Segments, UDP, doesn't seem like it has a defined term. Some people say segment, others say packet.

route-rs-runtime/src/packet/ipv4.rs
CollinValley: This is an old commit. In the latest I added a packet_offset parameter that means the packet can be agnostic to the amount of data preceeding it in the array.

route-rs-runtime/src/packet/ethernet.rs
CollinValley: /facepalm this is exactly what I wanted when I made the Result.  https://doc.rust-lang.org/std/convert/trait.TryFrom.html Will convert that over tonight.

route-rs-runtime/src/packet/ethernet.rs
CollinValley: Yeah, I will switch it over, similarly to how I did Ipv4/Ipv6. They keep an array internally.

route-rs-runtime/src/packet/ethernet.rs
CollinValley: I'll take a look into Error types. Ideally we could return formatted strings

CollinValley:
> I like this API. That said, it'd be much easier to read with constants, and I'd like to have a discussion about ways we can potentially make each inner packet independent of the outer packet storing it.

Yeah,  I made some strides on this front by retaining offsets as we go up stack.  A TCPsegment keeps track of the IP Header offset, it's own TCP Header offset, an then it's payload offset. This allows us to remain more agnostic about why layer is beneath us. Unfortunately, there are instances of coupling, like the combined checksum in TCP, that break the model.  Also the fact that TCP has no length field, it just uses the IP length. Super annoying.

route-rs-runtime/src/lib.rs
CollinValley: I want to agree, would be nice to version the library too, so I can tell users it is 0.1 lol.  Will create a patch to move the files.

route-rs-runtime/src/packet/types.rs
```diff
@@ -0,0 +1,427 @@
...
+    pub fn new(bytes: [u8; 6]) -> MacAddr {
+        MacAddr { bytes }
+    }
+}
+
+impl fmt::Display for MacAddr {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            " {:X}::{:X}::{:X}::{:X}::{:X}::{:X} ",
```
moosingin3space: Switch to single colon.

route-rs-runtime/src/packet/types.rs
```diff
@@ -0,0 +1,427 @@
...
+            self.bytes[3],
+            self.bytes[4],
+            self.bytes[5]
+        )
+    }
+}
+
+//Most significant byte is 0th
+#[derive(Eq, Clone, Copy, Hash, PartialEq, Debug)]
+pub struct Ipv4Addr {
```
moosingin3space: This already exists in the stdlib.

route-rs-runtime/src/packet/types.rs
```diff
@@ -0,0 +1,427 @@
...
+        write!(
+            f,
+            " {}.{}.{}.{} ",
+            self.bytes[0], self.bytes[1], self.bytes[2], self.bytes[3]
+        )
+    }
+}
+
+#[derive(Eq, Clone, Copy, Hash, PartialEq, Debug)]
+pub struct Ipv6Addr {
```
moosingin3space: See above.

route-rs-runtime/src/packet/ipv6.rs
```diff
@@ -0,0 +1,336 @@
...
+            | IpProtocol::HIP
+            | IpProtocol::Shim6
+            | IpProtocol::Use_for_expiramentation_and_testing => {
+                if data.len() <= offset + 1 {
+                    // Check for length overrun
+                    return IpProtocol::Reserved;
+                }
+                header_ext_len = data[offset + 1];
+                if header_ext_len == 0 {
+                    //fragments have the minimum of 8, but it set to zero for some dumb reason
```
moosingin3space: Can you cite your source here please?

route-rs-runtime/src/packet/ipv6.rs
```diff
@@ -0,0 +1,336 @@
...
+            .copy_from_slice(&addr.bytes()[..]);
+    }
+
+    pub fn set_dest_addr(&mut self, addr: Ipv6Addr) {
+        self.data[self.packet_offset + 24..self.packet_offset + 40]
+            .copy_from_slice(&addr.bytes()[..]);
+    }
+
+    //TODO: Test the get and set for extension headers.
+    pub fn extension_headers(&self) -> Vec<Cow<[u8]>> {
```
moosingin3space: Use a common function (between this and `get_ipv6_payload_type`)

moosingin3space (dismissed):

route-rs-runtime/src/packet/ipv6.rs
CollinValley: Took at look at this, while a good amount of the logic is the same. The main loop differs which makes abstracting the logic hard.

route-rs-runtime/src/packet/types.rs
CollinValley: Done

CollinValley:
I rebased this against master, so it once I have a little bit of approval we can merge this in. I think it is in a useful state right now, enough that I can call it version 0.1 It should also unblock us from building a router/elements using it.

route-rs-packets/src/ipv4.rs
```diff
@@ -0,0 +1,344 @@
...
+    pub fn set_checksum(&mut self) {
+        let new_checksum = self.caclulate_checksum();
+        self.data[self.packet_offset + 10] = ((new_checksum & 0xFF00) >> 8) as u8;
+        self.data[self.packet_offset + 11] = (new_checksum & 0x00FF) as u8;
+        self.validated_checksum = true;
+    }
+}
+
+/// Returns Ipv4 payload type, reads the header information to get the type
+/// of IpProtocol payload is included. Upon error, returns IpProtocol::Reserved.
```
eeverett6, line 225: Why not an error type? What if this reserved value is later assigned?

route-rs-packets/src/ipv6.rs
```diff
@@ -0,0 +1,352 @@
...
+        }
+        self.data.extend(payload);
+        if !headers.is_empty() {
+            self.set_next_header(first_header as u8);
+        }
+    }
+}
+
+/// Returns Ipv6 payload type, reads the header information to get the type
+/// of IpProtocol payload is included. Upon error, returns IpProtocol::Reserved.
```
eeverett6, line 184: Same question about reserved here

route-rs-packets/src/types.rs
```diff
@@ -0,0 +1,336 @@
...
+    HIP,
+    Shim6,
+    WESP,
+    ROHC,
+    Unassigned,
+    Use_for_expiramentation_and_testing,
+    Reserved,
+}
+
+impl From<u8> for IpProtocol {
```
eeverett6, line 197: Couple thoughts:
- It feels wordy to have to do this twice
- How do we turn an IpProtocol into a u8 when writing out a packet/overwriting the protocol field?

Have you seen the solution in libpnet? https://github.com/libpnet/libpnet/blob/master/pnet_packet/src/ip.rs It looks like the protocol is a struct with a single u8 field, then there are two mappings, one a list of constants, and one a display function.

route-rs-packets/src/ethernet.rs
```diff
@@ -0,0 +1,169 @@
...
+pub struct EthernetFrame<'packet> {
+    pub data: PacketData<'packet>,
+    pub payload_offset: usize,
+}
+
+impl<'frame> EthernetFrame<'frame> {
+    pub fn new(frame: PacketData) -> Result<EthernetFrame, &'static str> {
+        // Ethernet II frames must be at least the header, which is 14bytes
+        // 0                    6                    12                      14
+        // |---6 byte Dest_MAC--|---6 byte Src_MAC---|--2 Byte Payload Len---|
```
eeverett6: In Ethernet II, the 12th and 13th bytes are actually the ethertype field, not a payload length. I believe there is no length field in ethernet II. We will probably need something like the IpProtocol types for EtherTypes as well

route-rs-packets/src/ipv4.rs
```diff
@@ -0,0 +1,344 @@
...
+
+    pub fn checksum(&self) -> u16 {
+        u16::from_be_bytes(
+            self.data[self.packet_offset + 10..=self.packet_offset + 11]
+                .try_into()
+                .unwrap(),
+        )
+    }
+
+    pub fn dcsp(&self) -> u8 {
```
eeverett6: dscp

route-rs-packets/src/ethernet.rs
```diff
@@ -0,0 +1,169 @@
...
+    type Error = &'static str;
+
+    fn try_from(segment: UdpSegment<'packet>) -> Result<Self, Self::Error> {
+        EthernetFrame::new(segment.data)
+    }
+}
+
+impl<'packet> TryFrom<Ipv4Packet<'packet>> for EthernetFrame<'packet> {
+    type Error = &'static str;
+
```
eeverett6, line 80: Should we have a way of encapsulating an Ipv4 packet if it didn't originally come from an ethernet frame?

eeverett6 (commented):
Didn't have time to look over the whole thing tonight, but posting my feedback so far. I should be able to contribute commits to address some of this feedback myself soon too.

route-rs-packets/src/ipv4.rs
CollinValley: Idk making an error type is kind of annoying, this is quick and dirty. Error type is on the list of improvements, i can spawn off an issue.

CollinValley: I can fix this by returning a Result at least to start.

route-rs-packets/src/ipv6.rs
CollinValley: ack

route-rs-packets/src/types.rs
CollinValley: Hmm this is a decent solution.  I'd be fine with changing it to something like libpnet's

route-rs-packets/src/ipv4.rs
CollinValley: ack

route-rs-packets/src/ethernet.rs
CollinValley: In this case, this function creates an EthernetFrame from an Ipv4 or Ipv6.  So it isn't really encapsulating per se, just a type conversion.  We could add support for other Layer 2 protocols if need be, since the try_from would fail in that caseif you tried to make an ethernet frame.

route-rs-packets/src/ethernet.rs
CollinValley: Interesting.  https://en.wikipedia.org/wiki/Ethernet_frame Looks like it is EtherType in the case it is an Ethernet II frame. As far as I'm aware, we really only expect Ethernet II types on the internet?  https://en.wikipedia.org/wiki/EtherType Looks like it is fairly complex.

CollinValley:
Thanks for looking through it Ethan, obviously any contribution is welcome.  I'll start trying to address the simple things, and will create some issues for the stuff that will take longer.

route-rs-packets/src/ipv4.rs
moosingin3space: What kind of errors are possible?

route-rs-packets/src/types.rs
moosingin3space: Yeah, arguably `From<T>` isn't what we want here -- let's make clients explicitly convert integers to the typed values.

route-rs-packets/src/ethernet.rs
moosingin3space: EtherTypes are how we define what the next packet type is, though -- since LLDP and ARP don't use IP headers. I'm not sure we can avoid it. According to https://en.wikipedia.org/wiki/Ethernet_frame#Types, it looks like Ethernet II is the most common, with raw IEEE 802.3 being an ancient Novell thing.

route-rs-packets/src/types.rs
CollinValley: Going to shelve this for after merge cleanup.

CollinValley: #178

route-rs-packets/src/ipv4.rs
CollinValley: There are a few numbers that are undefined?

CollinValley: Edit: Nevermind this error makes sense. The only error is that they didn't hand me something that could have been an Ipv4 packet.

I am going to wrap this in a Result rather than a custom error to be more clear.

CollinValley:
Just resolved the last issue from the previous CRs. Also rebased the branch against master to ensure everything is still working as expected.

route-rs-packets/src/ethernet.rs
eeverett6: I would honestly just drop support for the 802.3 stuff entirely and just support Ethernet II with the EtherType field. I don't think it's worth it to be generic across so many different standards when we'll never encounter them.

route-rs-packets/src/ethernet.rs
eeverett6: I'm thinking of the case where say a processor/element creates an Arp Message, that now needs to be put behind an Ethernet header to be sent out over the wire. These would be separate from the TryFroms, but could be methods that take an arbitrary payload, a src/dest MAC addr, and an ethertype, then replaces the old l2 header or pushes more data onto the front of the packet as necessary

route-rs-packets/src/types.rs
```diff
@@ -0,0 +1,348 @@
...
+    Mobility_Header,
+    UDPLite,
+    MPLS_in_IP,
+    manet,
+    HIP,
+    Shim6,
+    WESP,
+    ROHC,
+    Unassigned,
+    Use_for_expiramentation_and_testing,
```
eeverett6: experimentation

route-rs-packets/src/ethernet.rs
CollinValley: We chatted offline about this, my next push will shift us to just supporting EtherType instead of length field. I will still check that it is in a valid EtherType range.

route-rs-packets/src/types.rs
CollinValley: whoops, fixed.

ddowl (approved):
lol I read this

PR #165, opened 2019-10-13
Switch to buildersyntax, create first composite

CollinValley:
Big PR,
Changes all of the links over to the new builder syntax. As well as does some renaming and housekeeping.


route-rs-runtime/src/element/identity_elements.rs
```diff
@@ -17,7 +17,10 @@ impl<A> IdentityElement<A> {
     }
 }
 
-impl<A> Element for IdentityElement<A> {
+impl<A> Element for IdentityElement<A>
```
ddowl: Can this not have the type constraint `A: Send`?

route-rs-runtime/src/link/async_link.rs
```diff
@@ -1,43 +1,98 @@
...
+        }
+    }
+}
+
+impl<E: Element + Send + 'static> LinkBuilder<E::Input, E::Output> for AsyncLinkBuilder<E> {
+    fn ingressors(self, mut in_streams: Vec<PacketStream<E::Input>>) -> Self {
+        assert_eq!(
+            in_streams.len(),
+            1,
+            "AsyncLink may only take 1 input stream"
```
ddowl, line 56: I may answer this later as I keep reading, but just confirming that we're keeping `Async` as 1 - 1, where we would use Link composition with `Join` and `Tee` in order to get the M - N effect?

route-rs-runtime/src/link/async_link.rs
```diff
@@ -214,56 +269,105 @@ impl<Packet: Sized> Stream for AsyncEgressor<Packet> {
...
+    use crate::element::{DropElement, IdentityElement, TransformElement};
     use crate::link::sync_link::SyncLinkBuilder;
-    use crate::link::{ElementLinkBuilder, LinkBuilder};
+    use crate::link::{ElementLinkBuilder, LinkBuilder, TokioRunnable};
     use crate::utils::test::packet_collectors::ExhaustiveCollector;
     use crate::utils::test::packet_generators::{immediate_stream, PacketIntervalGenerator};
     use core::time;
     use futures::future::lazy;
 
+    fn run_tokio(runnables: Vec<TokioRunnable>) {
```
ddowl, line 280: I'd like to consolidate these kinds of functions into test helpers, but I can do that after this PR is merged.

route-rs-runtime/src/link/async_link.rs
CollinValley: Exactly, which is why I changed it to queue link instead of async link

route-rs-runtime/src/element/identity_elements.rs
CollinValley: That seems reasonable, will change

route-rs-runtime/src/link/async_link.rs
```diff
@@ -214,56 +269,105 @@ impl<Packet: Sized> Stream for AsyncEgressor<Packet> {
...
-        let default_channel_size = 10;
         let packets = vec![0, 1, 2, 420, 1337, 3, 4, 5, 6, 7, 8, 9];
         let packet_generator = immediate_stream(packets.clone());
 
-        let elem = AsyncIdentityElement::new();
+        let elem = IdentityElement::new();
 
-        let link = AsyncLink::new(Box::new(packet_generator), elem, default_channel_size);
+        let (mut runnables, mut egressors) = AsyncLinkBuilder::new()
+            .ingressors(vec![Box::new(packet_generator)])
```
ddowl: For each of these test examples, you can use the `ingressor` function you defined. `immediate_stream` is actually already inside a `Box`, so you can just use  `.ingressor(packet_generator)`

route-rs-runtime/src/link/async_link.rs
CollinValley: I was thinking the same thing when I kept copying it lol

route-rs-runtime/src/link/async_link.rs
```diff
@@ -214,56 +269,105 @@ impl<Packet: Sized> Stream for AsyncEgressor<Packet> {
...
+        let identity_element1 = IdentityElement::new();
+
+        AsyncLinkBuilder::new()
+            .element(identity_element1)
+            .ingressor(packet_generator1)
+            .build_link();
+    }
+
     #[test]
     fn async_link() {
```
ddowl: nit: This test name should be more descriptive, or we can remove the example if it's not covering anything useful.

route-rs-runtime/src/link/join_link.rs
```diff
@@ -1,51 +1,75 @@
...
+}
 
-            let ingressor =
-                JoinIngressor::new(input_stream, to_egressor.clone(), Arc::clone(&task_park));
-            ingressors.push(ingressor);
-            from_ingressors.push(from_ingressor);
-            task_parks.push(task_park);
+impl<Packet: Sized + Send + 'static> LinkBuilder<Packet, Packet> for JoinLinkBuilder<Packet> {
+    fn ingressors(self, in_streams: Vec<PacketStream<Packet>>) -> Self {
+        assert_ne!(in_streams.len(), 0, "Input stream vector can not be zero!");
```
ddowl: Should we assert a rough upper bound on the number of ingress streams for sanity? I can't imagine a good design having more than 20 or so.

route-rs-runtime/src/link/tee_link.rs
```diff
@@ -1,50 +1,103 @@
...
+
+    pub fn branches(self, branches: usize) -> Self {
         assert!(
             branches <= 1000,
             format!("Tee Link branches: {} > 1000", branches)
         );
-        assert!(
-            queue_capacity <= 1000,
-            format!("Tee Link queue_capacity: {} > 1000", queue_capacity)
+        assert_ne!(branches, 0, "branches must be non-zero");
```
ddowl: Similar to `Join`, should we restrict the number of outgoing branches from `Tee`?

route-rs-runtime/src/link/tee_link.rs
```diff
@@ -1,50 +1,103 @@
...
+
+        TeeLinkBuilder {
+            in_stream: self.in_stream,
+            queue_capacity: self.queue_capacity,
+            branches: Some(branches),
+        }
+    }
+}
+
+impl<Packet: Sized + Send + Clone + 'static> LinkBuilder<Packet, Packet>
```
ddowl: Eventually the trait constraints on what can be passed through `Link`s might get a little out of hand, so maybe later we should alias those into another trait:

```
trait LinkPassable: Sized + Send + Clone + 'static
```

so we can write this definition as

```
impl<P: LinkPassable> LinkBuilder<P, P>
```

route-rs-runtime/src/link/tee_link.rs
```diff
@@ -1,50 +1,103 @@
...
+#[derive(Default)]
+pub struct TeeLinkBuilder<Packet: Sized + Clone + Send> {
+    in_stream: Option<PacketStream<Packet>>,
+    queue_capacity: usize,
+    branches: Option<usize>,
 }
 
-impl<P: Sized + Clone> TeeLink<P> {
-    pub fn new(input_stream: PacketStream<P>, queue_capacity: usize, branches: usize) -> Self {
+impl<Packet: Sized + Clone + Send> TeeLinkBuilder<Packet> {
```
ddowl: We could also define an `ingressor` method for `TeeLink` that works similarly to other `Link`s that take just one ingressor.

route-rs-runtime/src/composite/mton_composite.rs
```diff
@@ -0,0 +1,157 @@
+use crate::link::{Link, LinkBuilder, PacketStream, JoinLinkBuilder, TeeLinkBuilder};
+
+pub struct  MtoNCompositeBuilder<Packet: Sized + Send + Clone> {
+    in_streams: Option<Vec<PacketStream<Packet>>>,
+    join_queue_capacity: usize,
+    tee_queue_capacity: usize,
+    branches: Option<usize>,
```
ddowl: Is there a better name than `branches` to express the number of egressors the `Link` will produce? How about `num_egressors`?

route-rs-runtime/src/composite/mton_composite.rs
```diff
@@ -0,0 +1,157 @@
...
+                .ingressors(self.in_streams.unwrap())
+                .queue_capacity(self.join_queue_capacity)
+                .build_link();
+            let (mut tee_runnables, tee_egresssors) = TeeLinkBuilder::new()
+                .ingressors(join_egressors)
+                .queue_capacity(self.tee_queue_capacity)
+                .branches(self.branches.unwrap())
+                .build_link();
+            tee_runnables.extend(join_runnables);
+            (tee_runnables, tee_egresssors)
```
ddowl: nit: Could we use `Vec::append` rather than `extend` to make it clear that we're returning all the runnables? I'd even prefer an extra name like:

```
let all_runnables = tee_runnables.append(join_runnables);
(all_runnables, tee_egressors)
```

route-rs-runtime/src/composite/mod.rs
```diff
@@ -0,0 +1,4 @@
+/// Basic composite that take in M streams of a type, and outputs that same
+/// type to N packets.
+mod mton_composite;
```
ddowl, line 3: nit: Since the job of this link isn't to really transform anything, just take M streams of a type and output copies to N streams, can we call it `MultiTee` or `MultiClone` or something? `composite` doesn't really mean anything on its own, and in fact we might want to hide the fact that any given Link is a composite since that's an implementation detail.

route-rs-runtime/src/lib.rs
```diff
@@ -26,3 +26,7 @@ pub mod pipeline;
 
 /// Utility module
 mod utils;
+
+/// Composites are groups of links pre-assmebled to provide higher level functionality. They are highly customizable and users of the
+/// library are encourged to make their own to encourage code reuse.
```
ddowl, line 31: :100:

route-rs-runtime/src/link/mod.rs
```diff
@@ -12,8 +12,8 @@ pub use self::sync_link::*;
 /// useful for creating queues in the router, buffering, and creating `Task` boundries that can be processed on
 /// different threads, or even different cores. Before packets are placed into the queue to be output, they are run
 /// through the element defined process function, often performing some sort of transformation.
-mod async_link;
-pub use self::async_link::*;
+mod queue_link;
```
ddowl, line 15: nit: I know that the queue structure is exposed via builder method, but if I'm a user I'm not quite sure what the "queue" gets me. This link can output to multiple egressors and wait for long-running processes to complete (which `SyncLink` can't),  but that's not obvious from the name `QueueLink`. I'm also not a big fan of `AsyncLink` either though. I'm cool with this for now, but we should definitely think hard about the names of our primitive links before any kind of release. We might notice some more API UX problems like, "if there's a SyncLink why is there no AsyncLink?", that might be more noticeable after finishing the NAT and having other people play around with the code. _user interviews??_

To put some more names into the ring: `BufferedLink` and `UnbufferedLink`. They're complements, and you can kindof intuit their properties based on the names?

route-rs-runtime/src/link/classify_link.rs
```diff
@@ -1,64 +1,140 @@
...
+            branches: self.branches,
+        }
+    }
+
+    pub fn branches(self, branches: usize) -> Self {
         assert!(
-            queue_capacity <= 1000,
-            format!("ClassifyLink queue_capacity: {} > 1000", queue_capacity)
+            (2..=1000).contains(&branches),
+            format!("Branches: {}, must be in range 2..=1000", branches)
```
ddowl: "must be between 2 and 1000 branches"?

route-rs-runtime/src/link/classify_link.rs
```diff
@@ -1,64 +1,140 @@
...
 }
 
-impl<'a, C: Classifier> ClassifyLink<'a, C> {
-    pub fn new(
-        input_stream: PacketStream<C::Packet>,
-        classifier: C,
-        dispatcher: Box<dyn Fn(C::Class) -> usize + Send + Sync + 'a>,
-        queue_capacity: usize,
-        branches: usize,
+impl<C: Classifier> ClassifyLink<C> {
```
ddowl, line 19: We can also define an `ingressor` function on `ClassifyLink` that accepts only 1 `PacketStream`.

route-rs-runtime/src/link/classify_link.rs
```diff
@@ -167,37 +244,97 @@ mod tests {
...
 
-        let mut even_link = ClassifyLink::new(
-            packet_generator,
-            even_classifier,
-            Box::new(|evenness| if evenness { 0 } else { 1 }),
-            default_channel_size,
-            number_branches,
-        );
+        let (mut runnables, mut egressors) = ClassifyLink::new()
+            .ingressors(vec![Box::new(packet_generator)])
```
ddowl: `ClassifyLink`'s test examples can also use the simpler `ingressor` builder function as well.

route-rs-runtime/src/link/mod.rs
```diff
@@ -34,9 +34,9 @@ pub use self::blackhole_link::*;
 use crate::element::Element;
 
 /// All Links communicate through streams of packets. This allows them to be composable.
-pub type PacketStream<Input> = Box<dyn Stream<Item = Input, Error = ()> + 'static + Send>;
+pub type PacketStream<Input> = Box<dyn Stream<Item = Input, Error = ()> + Send>;
 /// Some Links may need to be driven by Tokio. This represents a handle to something Tokio can run.
-pub type TokioRunnable = Box<dyn Future<Item = (), Error = ()> + 'static + Send>;
+pub type TokioRunnable = Box<dyn Future<Item = (), Error = ()> + Send>;
```
ddowl: Is removing the static lifetime on these types intentional?

route-rs-runtime/src/link/async_link.rs
CollinValley: I'm not sure that we should have an `ingressor` function at all. I think it kind of breaks our own trait concept? In a similar vein, for other functions, like braches, queue_length, etc, should we pull those into the trait and then make the links that don't need them implement them as no-ops?

route-rs-runtime/src/link/tee_link.rs
CollinValley: Yeah, something like that.  I  want to talk about this in person. Since one we can't make packets 'static, and two we should reevaluate what our Trait requirements are in general.

route-rs-runtime/src/link/tee_link.rs
CollinValley: I feel like it is kludgy to have special methods in general for certain links,  would like to discuss with the group in person, see what everyone thinks.

route-rs-runtime/src/composite/mton_composite.rs
CollinValley: ack, good suggestion.

route-rs-runtime/src/link/async_link.rs
ddowl: I see it as a convenience method that isn't part of the `Link` trait. Something that users can specify when creating their own HoL (higher-order links) or composites, but obviously something we can't do in graphgen. As long as we're sure that

```builder.ingressor(in) == builder.ingressors(vec![Box::new(in)])```

then nothing is broken, and users get a slighly more ergonomic method to use.

File: route-rs-runtime/src/composite/mton_composite.rs (hunk 11 above)
Comment (CollinValley) [reply route-rs-runtime/src/composite/mton_composite.rs]: I suppose, I don't object to append.

File: route-rs-runtime/src/link/tee_link.rs (hunk 7 above)
Comment (CollinValley) [reply route-rs-runtime/src/link/tee_link.rs]: Seems reasonable

File: route-rs-runtime/src/composite/mod.rs (hunk 12 above)
Comment (CollinValley) [reply route-rs-runtime/src/composite/mod.rs:3]: Hmm I like MultiClone, or some other spiffy name, I agree the composite it kind of an impl detail.

File: route-rs-runtime/src/link/mod.rs (hunk 14 above)
Comment (CollinValley) [reply route-rs-runtime/src/link/mod.rs:15]: Hmm,  I do agree not having a Sync/Async complement is an issue. Not sure I like buffered link,   FYR, the queuelink is still 1-1.  If you want multi-output, then you need a CloneLink.   I thought about supporting the queuelink as multi-output, but figured it would be better to just put a SyncLink->CloneLink to get the desired effect at minimal cost.  

Crazytown thought. What if the only Link that can take elements is the SyncLink?  QueueLink or BufferLink is for...buffers?  We could rename Clone to Fork, so there are obvious Fork/Join complements.  Classify would remain its own special link

Comment (ddowl) [reply route-rs-runtime/src/link/mod.rs:15]: https://gph.is/g/E3geJNn

**Comment by CollinValley** [comment]
Recent changes.

- Resolved weirdness with Asserts, they are all much more consistent
- BlackholeLink: Tried to eliminate the 'static packet,  by requiring a phantom Element.  This is less than ideal, but I think it works to fix our issue.  If we go down this route, we would likely make the element real rather than a dummy.
- 'static lifetime: Fixed the error where I removed the 'static requirement from TokioRunnable, left it off on PacketStream?  It doesn't seem to cause an issue.

More work still to do, but this is what I could do on the plane with no internet.

File: route-rs-runtime/src/link/async_link.rs (hunk 4 above)
Comment (CollinValley) [reply route-rs-runtime/src/link/async_link.rs]: Hmm, put that way,  I do like having it there.  Might form a nice marker for the user to know that the link is 1 to stuff.

File: route-rs-runtime/src/link/join_link.rs (hunk 6 above)
Comment (CollinValley) [reply route-rs-runtime/src/link/join_link.rs]: I set this to 1000, decided to let people be dumb if they want. Left it at 1000 for ingress streams, egressors, and queue length. I doubt anybody wants numbers this high for anything else.

File: route-rs-runtime/src/link/tee_link.rs (hunk 7 above)
Comment (CollinValley) [reply route-rs-runtime/src/link/tee_link.rs]: Fixed

File: route-rs-runtime/src/composite/mton_composite.rs (hunk 10 above)
Comment (CollinValley) [reply route-rs-runtime/src/composite/mton_composite.rs]: Done

File: route-rs-runtime/src/link/mod.rs (hunk 14 above)
Comment (CollinValley) [reply route-rs-runtime/src/link/mod.rs:15]: Going to leave this thought for meetup. We should make the choice as a group, since it basically will cement the design philosophy for our links overall.

File: route-rs-runtime/src/link/classify_link.rs (hunk 15 above)
Comment (CollinValley) [reply route-rs-runtime/src/link/classify_link.rs]: At first I thought I would protect the user from making a 1-1 classify cause what is the point? Same for 1-1 CloneLink. Decided to remove that cause it isn't completely the libraries job to protect the user from stupid inefficiency. There may be a use we aren't considering.

File: route-rs-runtime/src/link/mod.rs (hunk 18 above)
Comment (CollinValley) [reply route-rs-runtime/src/link/mod.rs]: Sorta, patched this.

File: route-rs-runtime/src/element/identity_elements.rs (hunk 1 above)
Comment (CollinValley) [reply route-rs-runtime/src/element/identity_elements.rs]: done

File: route-rs-runtime/src/link/async_link.rs (hunk 5 above)
Comment (CollinValley) [reply route-rs-runtime/src/link/async_link.rs]: doneish

File: route-rs-runtime/src/link/tee_link.rs (hunk 8 above)
Comment (CollinValley) [reply route-rs-runtime/src/link/tee_link.rs]: I did go through and remove the Sized requirement, since Clone requires Sized.

File: route-rs-runtime/src/link/async_link.rs (hunk 4 above)
Comment (CollinValley) [reply route-rs-runtime/src/link/async_link.rs]: Went ahead and implemented this for all links. On links that take multiple inputs, it works like a push() function. One links that only take one input, works as replace()

File: route-rs-runtime/src/link/tee_link.rs (hunk 9 above)
Comment (CollinValley) [reply route-rs-runtime/src/link/tee_link.rs]: Done

File: route-rs-runtime/src/link/classify_link.rs (hunk 16 above)
Comment (CollinValley) [reply route-rs-runtime/src/link/classify_link.rs:19]: done

File: route-rs-runtime/src/link/classify_link.rs (hunk 17 above)
Comment (CollinValley) [reply route-rs-runtime/src/link/classify_link.rs]: done

File: route-rs-runtime/src/composite/mton_composite.rs (hunk 11 above)
Comment (CollinValley) [reply route-rs-runtime/src/composite/mton_composite.rs]: done

File: route-rs-runtime/src/link/queue_link.rs [hunk 19]
```diff
@@ -327,7 +327,7 @@ mod tests {
     }
 
     #[test]
-    fn queue_link() {
+    fn queue_link_works() {
```
ddowl, line 330: HAHAHAHA

ddowl:
LGTM. Let's try and make PRs a little smaller in the future.

ddowl (approved):
Whoops, meant to approve.

PR #174, opened 2019-10-25
Rename links, Organize runtime entities

### LINKED ISSUE #170: Link Rename
Blackhole -> Drop
Clone -> Fork
Sync -> Process

Element -> Processor

ddowl:
This hefty change resolves #170 , renaming `Link`s based on some bikeshedding conversation at the last meetup. It also performs some other module-level restructuring such as (1) creating a module for the `Classifier` trait, implementations, and link builders (examples are used in some link tests), and (2) creating 3 submodules of `route-rs-runtime::link`: `composite`, `primitive`, and `utils`. Those commits are prefixed with "structure", wheras renaming commits are prefixed with "rename" vaguely following [conventional commit style](https://www.conventionalcommits.org/en/v1.0.0/). It might be easier to review this PR by-commit.

Although this change is a functional no-op (our test suite is awesome :100:), some docs might look strange since find-replace did most of the work. Since we need to review the docs anyways, comment on any wording weirdness or out of date material you come across. Or it might be easier to create a separate PR to update docs since it's orthogonal to renaming/restructuring, so at least take mental notes.

A potential extension to this PR could be to drop the `Link` suffix from `Link` objects altogether. _"Whoa, slow down there partner!"_ I know it's a bit radical, but it might be more clear to refer to refer to links with their namespaces. For comparison, I dropped `Processor` from the names of things that `impl Processor`. This gives somewhat cleaner imports:
```
    use crate::processor::{Drop, Identity, TransformFrom};
```
and arguably better builder syntax:
```
...
    .processor(Identity::new())
...
```
or with a more qualified name
```
...
    .processor(processor::Identity::new())
...
```
If it's too confusing we don't have to remove the suffixes from `Processor`s or `Link`s, but it's interesting to think about if we can use the context from module structure and the type system to give us more clear/concise names. I mean _of course_ the `.processor` builder function takes a `Processor` so why add the extra suffix for clarity if the function type already supplies that context?

@scgruber I let the find-replace into the graphgen. I hope it's helping more than hurting :sweat_smile:


route-rs-runtime/src/link/composite/m_transform_n_composite.rs
```diff
@@ -1,5 +1,6 @@
 use crate::link::{
-    ForkLink, JoinLink, Link, LinkBuilder, PacketStream, ProcessLink, ProcessLinkBuilder,
+    primitive::{ForkLink, JoinLink, ProcessLink},
```
CollinValley: I had no idea this worked...but _of course it does_

CollinValley (approved):
LGTMMD, good cleanup.

PR #175, opened 2019-10-27
Convert channel links

scgruber:
This PR converts InputChannelLink and OutputChannelLink to the new LinkBuilder style. It also fixes the examples to be compliant with the new style. I'm not updating graphgen in this PR because this already feels substaintial.

Marked as Draft for now because I haven't finished OutputChannelLink.


examples/trivial-identity/src/pipeline.rs
```diff
@@ -17,15 +18,25 @@ impl route_rs_runtime::pipeline::Runner for Pipeline {
...
+            .ingressor(Box::new(link_1))
+            .processor(elem_1_identity)
+            .build_link();
+        all_runnables.append(&mut runnables_2);
+        let link_2_egress_0 = link_2.remove(0);
 
-        let link_3 = OutputChannelLink::new(Box::new(link_2), output_channel);
+        let link_3 = OutputChannelLink::new(link_2_egress_0, output_channel);
 
         tokio::run(lazy(move || {
```
ddowl, line 42: We should have a common function for "run all tasks on tokio".

examples/dns-interceptor/src/pipeline.rs
```diff
@@ -17,43 +18,53 @@ impl route_rs_runtime::pipeline::Runner for Pipeline {
         input_channel: crossbeam::Receiver<Self::Input>,
         output_channel: crossbeam::Sender<Self::Output>,
     ) {
+        let mut all_runnables: Vec<TokioRunnable> = vec![];
+
         let elem_1_setinterfacebydestination = SetInterfaceByDestination::new();
         let elem_2_classifydns = ClassifyDNS::new();
         let elem_3_localdnsinterceptor = LocalDNSInterceptor::new();
 
         let link_1 = InputChannelLink::new(input_channel);
```
ddowl: It would be nice if the links had more descriptive names, but if this is a code generation limitation then it's fine.

examples/dns-interceptor/src/pipeline.rs
```diff
@@ -17,43 +18,53 @@ impl route_rs_runtime::pipeline::Runner for Pipeline {
...
 
-        let link_2 = ProcessLink::new(Box::new(link_1), elem_1_setinterfacebydestination);
+        let (mut runnables_2, mut link_2) = ProcessLink::new()
```
ddowl: The second tuple element is the egressors of the built `Link`.

route-rs-runtime/src/link/primitive/input_channel_link.rs
```diff
@@ -0,0 +1,98 @@
...
+
+    #[test]
+    fn immediate_packets() {
+        let packets = vec![0, 1, 2, 420, 1337, 3, 4, 5, 6, 7, 8, 9];
+        let (send, recv) = crossbeam_channel::unbounded();
+
+        let link = InputChannelLink::new().channel(recv).build_link();
+
+        for p in packets.clone() {
+            match send.send(p) {
```
ddowl, line 88: lol `send.send()`. Maybe we can call the `Sender` and `Receiver` `(s, r)`? That's what I've seen in other crossbeam examples.

examples/trivial-identity/src/pipeline.rs
scgruber: Yeah I think I'll add this as a separate PR when I'm working on graphgen

examples/dns-interceptor/src/pipeline.rs
scgruber: I think it's sort of a question of how we expect people to be working with the pipeline. Right now the Links are kind of invisible, and what you see is the Processors and such. I do think we could do more things to make the file navigable in big projects.

route-rs-runtime/src/link/primitive/input_channel_link.rs
```diff
@@ -0,0 +1,98 @@
...
+        panic!("InputChannelLink does not take stream ingressors")
+    }
+
+    fn build_link(self) -> Link<Packet> {
+        if self.channel_receiver.is_none() {
+            panic!("Cannot build link! Missing channel");
+        } else {
+            (
+                vec![],
+                vec![Box::new(StreamFromChannel {
```
CollinValley, line 35: This is pretty nifty.  Makes sense that this would only be an egressor, and has no tokio-runnable cost.

route-rs-runtime/src/link/primitive/output_channel_link.rs
```diff
@@ -0,0 +1,170 @@
...
+impl<Packet> Future for StreamToChannel<Packet> {
+    type Item = ();
+    type Error = ();
+
+    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+        loop {
+            if self.channel_sender.is_full() {
+                // Since we don't know anything about the other side of our channel, we have to
+                // self-notify and just hope that the other side empties it eventually.
+                task::current().notify();
```
CollinValley, line 76: This is expensive to do. It is 100% better to do than a tight spin loop, and I understand why we do it. But this means we should only making a habit of using the egressor-channel interface sparingly.

Ideally we would have some way to subscribe ourselves to be awoken by something in this interface, something to watch out for when we go try to plug this in to whatever is sending the packets to the interface.

CollinValley (dismissed):
Overall this looks good to me. 

I think it is important we don't use this a lot cause it is very greedy with the self-notify. But it is fit for purpose.

You may be able to use this to replace all our dedicated collectors in the testing harness. It is essentially the same ish.

CollinValley (dismissed):
LGTMMD

ddowl (approved):
LGTMMMFD

PR #176, opened 2019-10-28
Remove primitive DropLink. Create DropLink composite with random weighted distribution support.

ddowl:
Specifically, modifies the `Drop` processor to use a Bernoulli distribution to determine which packets get dropped, and moves the `DropLink` to the `composite` directory since it is now a composition of a `ProcessLink` with a `Drop` processor. I've restricted the number of ingress streams of this link to 1, since I don't want users to pay an extra cost of merging those ingress streams using a `JoinLink` unless they need to. 

It's worth noting that this implementation is _slightly_ less efficient in the always-drop case since we need to sample a Bernoulli distribution for every packet, but it's unclear if this will make a difference. If we _need_ a more efficient multi-ingress pure-drop `Link`, we can re-introduce the previous version of `DropLink` with its own tokio task, but since that doesn't seem necessary we should favor `Link` composition.

### DIFF
--- a/route-rs-runtime/src/link/composite/drop_link.rs
+++ b/route-rs-runtime/src/link/composite/drop_link.rs
@@ -100,11 +100,11 @@ mod tests {
     fn finishes() {
         let packets = vec![0, 1, 2, 420, 1337, 3, 4, 5, 6, 7, 8, 9];
 
-        let link: Link<i32> = DropLink::new()
+        let link = DropLink::new()
             .ingressor(immediate_stream(packets.clone()))
             .build_link();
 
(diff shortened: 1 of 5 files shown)


CollinValley:
I'm a bit confused by this PR.  I may be missing something, but how is DropLink special compared to ProcessLink with a specific Processor? This also changed drop link to a thing that may have an egressor? But doesn't in some cases?  My specific question is why can't you create a Processor that drops x% of packets through it, then just load it into your favorite ProcessorLink?

I get the concept of the change, but it seems like it could be better implemented as, not a link essentially.

CollinValley (approved):
LGTMMD

PR #177, opened 2019-10-29
afpacket: change unit test to be marked as ignored

moosingin3space:
This way, a bare `cargo test` invocation will run all the tests
excluding the `afpacket` crate tests. In order to actually include the
`afpacket` tests, you'll have to run `cargo test -- --ignored`.


scgruber (approved):
Approved, assuming the CI actually passes

PR #180, opened 2019-10-30
Initial documentation website

bweis:
The hugo source is in the `/docs` folder.  Deployed by running `publish_to_ghpages.sh` from the root directory.  Ideally, we would be able to run this script automagically once a branch is merged into master.  I will leave that to someone smarter than me though.


ddowl (approved):
Cool! We should think about how we want to present our GH README and link to the greater docs, but that can be a separate PR.

CollinValley (approved):
LGTMMD

Awesome work, makes us look a lot more professional

PR #181, opened 2019-11-05
Remove LLVM-exception from deny.toml

scgruber:
In cargo-deny 0.3.0, it is no longer valid to include the exception as
a separate line item. Since our checker passes without this line, let's
remove it.

Additionally, let's be more explicit about the version of cargo-deny
that is installed so we don't get surprises in the future.


ddowl (approved):
Thanks for digging into this @scgruber . Approved.

PR #182, opened 2019-11-06
Convert graphgen to use syn in many places and fix example generation

scgruber:
This PR will fix graphgen to be able to generate the current version of the examples. Along the way, convert a lot of the code generation to use the `syn` crate, so we don't have to implement the AST or string conversions ourselves.

Eventually, graphgen should just produce one big `syn::File` object and then write it to the file, but for now we can generate some pieces independently and still stick them together.


CollinValley (approved):
Looks good, the Syn crate really improves the structure of the generation, it feels a lot more precise. Seems like it is easier to make small adjustments too without causing too many side effects. 
+2 LGTMMD

PR #185, opened 2019-11-16
Minimal static router

CollinValley:
This is a basic implementation of a static router. It does not do much other than decide on an interface to send ip packets out of, based on some static settings in the implemented classifiers. I think it is a good first example for now. This pull request also includes changes to Packets that needed to be made to support this implementation. I think they are good changes that were required to make the Packets library usable in our runtime.  

After this PR, I envision we actually start work on "The Great Example", a fully functional router than could be connected to the internet and used to route traffic.  Static routing is a small but important component of that, and we can use the  lessons learned here to educate our efforts in the future.


examples/minimal-static-router/src/main.rs
```diff
@@ -0,0 +1,195 @@
...
+
+mod classifiers;
+mod processors;
+
+fn main() {
+    let data_v4: Vec<u8> = vec![
+        0xde, 0xad, 0xbe, 0xef, 0xff, 0xff, 1, 2, 3, 4, 5, 6, 8, 00, 0x45, 0, 0, 20, 0, 0, 0, 0,
+        64, 17, 0, 0, 192, 178, 128, 0, 10, 0, 0, 1,
+    ];
+    let data_v6: Vec<u8> = vec![
```
ddowl, line 16: nit: It isn't obvious what these bytes mean. Our examples could be a bit more descriptive if we can build up `EthernetFrame`s from an empty one.

route-rs-runtime/src/lib.rs
```diff
@@ -27,4 +27,4 @@ pub mod link;
 pub mod pipeline;
 
 /// Utility module
-mod utils;
+pub mod utils;
```
ddowl, line 30: note: I intended for these utilities (namely test utilities) were supposed to be private to the runtime, but it makes sense to export them so users can run tests on their router graphs.

ddowl (approved):

examples/minimal-static-router/src/main.rs
CollinValley: Good idea, I think a builder syntax or something is the next concept to support for Packets. I think doing that will allow us to have descriptive tests. Also Eric has his Pcap inputter done so we can leverage that as well in test cases.

route-rs-runtime/src/lib.rs
CollinValley: Yeah..we will go over all of our ways to run a router this next meetup and refactor it into one coherent tool

PR #188, opened 2019-11-19
Docs fixup

CollinValley:
Starting to create baseline documentation around our library. View this as a rough draft of the docs, but they are mildly usable now in the runtime after this PR

As a reminder you can build the local docs with
`cargo docs --no-deps --open`

### DIFF
--- a/route-rs-runtime/src/classifier/mod.rs
+++ b/route-rs-runtime/src/classifier/mod.rs
@@ -2,7 +2,7 @@
 //!
 //! Classifiers are very similar to processors, but are used to differentiate a stream of packets. As such, they take each packet by reference,
 //! and are not able to modify it. They are only used in the ClassifyLink. Classifiers are able to return any type, but generally return an Enum
-//! that will enform the Dispatch section of the ClassifyLink which group each packet belongs to. The Dispatch then moves each packet to a port
+//! that will inform the Dispatch section of the ClassifyLink which group each packet belongs to. The Dispatch then moves each packet to a port
 //! based on its classification.
 mod even;
 pub use self::even::*;
--- a/route-rs-runtime/src/classifier/mod.rs
+++ b/route-rs-runtime/src/classifier/mod.rs
@@ -1,3 +1,9 @@
+//! # What are they for?
+//!
+//! Classifiers are very similar to processors, but are used to differentiate a stream of packets. As such, they take each packet by reference,
+//! and are not able to modify it. They are only used in the ClassifyLink. Classifiers are able to return any type, but generally return an Enum
+//! that will enform the Dispatch section of the ClassifyLink which group each packet belongs to. The Dispatch then moves each packet to a port
+//! based on its classification.
 mod even;
 pub use self::even::*;
 
--- a/route-rs-runtime/src/lib.rs
+++ b/route-rs-runtime/src/lib.rs
@@ -1,30 +1,21 @@
+//! In brief: The runtime is what takes the basket of computation required by the user, links it together into the desired
+//! graph, and preps it to be handed to the desired runtime for running.
 #[macro_use]
 extern crate futures;
 extern crate crossbeam;
 extern crate tokio;
 
-/// Processors are the unit of transformation in route-rs. The Processor is defined by a trait that requires that
-/// creators of processors implement a process function, that the underlying link will call on every packet that
-/// moves through it. You can use processors to classify, modify headers, count packets, update state, create new packets,
-/// join packets, whatever you want! As long as an processor conforms to the trait specification, the processor can be run inside a route-rs.
-/// While there are many provided processors that can be used to implement a router, users of route-rs that need specifc functionality
-/// in their router most likely will implement their own custom processors, conforming to the laid out processor standard.
+/// Implement the basic transformations for packets in the router.
 pub mod processor;
 
+/// Implement classification functions, helping the router decide to send a packet down route A,B,C..and so on.
 pub mod classifier;
 
-/// Links are an abstraction used by the runtime to link processors together, and manage the flow of packets through the router. Processors, which
-/// implement all the non-flow business logic of the router, are loaded into links to create specfic behavior. Links are joined together via their
-/// interfaces, and the links are then dumped into a runtime to being pulling packets through the router. In general, users of route-rs are not
-/// expected to have to implement their own links, because all of the desired flows should already be representable with the provided selection.
-/// Links are usually declared, connected, and placed into a runtime via generated code, where the user lays out their desired graph of processors
-/// and our graphgen program creates the pipeline that defines the desired router. If you are looking for a place to start, start there.
+/// Wrappers around Processors and Classfiers, and implement all the movement of Packets through the Router.
 pub mod link;
 
-/// Pipelines are abstractions used by graphgen to define a grouping of links that are run by the runtime, in our case Tokio. The graphgen
-/// program takes the user provided graph, and generates a pipeline. It connects input and output channels to the pipe, so that it can be
-/// hooked up to packet sources and sinks, and drops the pipeline into the runtime. Generally can be abstracted to the concept, router.
+/// Structure that defines a complete router, with input and output interfaces that may be hooked into the host system.
 pub mod pipeline;
 
-/// Utility module
+/// Utilities for the Runtime. Mostly testing constructs.
 pub mod utils;
--- a/route-rs-runtime/src/lib.rs
+++ b/route-rs-runtime/src/lib.rs
@@ -1,5 +1,5 @@
 //! In brief: The runtime is what takes the basket of computation required by the user, links it together into the desired
-//! graph, and preps it to be handed to the desired runtime for running.
+//! graph, and preps it to be handed to the Tokio for running.
 #[macro_use]
 extern crate futures;
 extern crate crossbeam;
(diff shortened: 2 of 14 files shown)


route-rs-runtime/src/lib.rs
```diff
@@ -1,30 +1,21 @@
+//! In brief: The runtime is what takes the basket of computation required by the user, links it together into the desired
+//! graph, and preps it to be handed to the desired runtime for running.
```
ddowl: The desired runtime being Tokio? Or are we saying that the async runtime should be configurable?

route-rs-runtime/src/classifier/mod.rs
```diff
@@ -1,3 +1,9 @@
+//! # What are they for?
+//!
+//! Classifiers are very similar to processors, but are used to differentiate a stream of packets. As such, they take each packet by reference,
+//! and are not able to modify it. They are only used in the ClassifyLink. Classifiers are able to return any type, but generally return an Enum
+//! that will enform the Dispatch section of the ClassifyLink which group each packet belongs to. The Dispatch then moves each packet to a port
```
ddowl: > enform

inform?

route-rs-runtime/src/link/mod.rs
```diff
@@ -1,14 +1,29 @@
+//! # What are they for?
+//!
+//! Links are an abstraction used by the runtime to link processors and classifiers together, and manage the flow of packets through the router.
+//! Processors and Classifiers, which implement all the non-flow business logic of the router, are loaded into links to create specfic behavior.
+//! Links are joined together via their interfaces, and the links are then dumped into a runtime to begin pulling packets through the router.
```
ddowl: > links are then dumped into a runtime to begin pulling packets...

We can do a better job of explaining what the runtime is for here. How about something like:
> Links are composed together via the `Link` trait and handed to Tokio (some Links, such as `QueueLink`, are async and rely on Tokio to pull packets from upstream).

route-rs-runtime/src/link/mod.rs
```diff
@@ -1,14 +1,29 @@
...
+//! When the library user uses Graphgen, Links are declared, connected, and placed into a runtime via generated code. The user does this by laying
+//! out their desired graph of Processors and Classifiers in a graph file, and Graphgen does all the work from there! Additionally, since all links
+//! must implement the LinkBuilder trait, Links are transparently composable at no cost during runtime! The most basic links that implement Poll,
+//! Stream and Future are Primitives. Links that are made by wrapping a collection of Links are called Composites. Link composition is a great way
+//! to reduce the complexity of your router, and make its design more inspectable. Users of the library are encouraged to create their own Composite
+//! Links, but should refrain from defining new Primitive Links. This prevents the user from having to worry about the complexities of generically
```
ddowl: > ...should refrain from defining new Primitive Links.

We can remove this sentence if we disallow users from adding links to the same Tokio instance?

route-rs-runtime/src/link/mod.rs
```diff
@@ -1,14 +1,29 @@
...
+
 use crate::processor::Processor;
 use futures::{Future, Stream};
 
 /// Composites are groups of links pre-assmebled to provide higher level functionality. They are highly customizable and users of the
 /// library are encourged to make their own to encourage code reuse.
 pub mod composite;
 
-// TODO: primitives docs
+/// Primitive links are links that individually implement all the flow based logic that can be combined to create more compilcated
```
ddowl: Can remove "are links that"

route-rs-runtime/src/link/mod.rs
```diff
@@ -1,14 +1,29 @@
...
+/// composite links. Users of the library should not have to implement their own primitive links, but should rather combine them into
+/// their own custom composite links.
 pub mod primitive;
 
-// TODO: util docs
+/// Commmon utilities used by links, for instance the `task_park` utility used in primitive links to facilite sleeping and waking.
```
ddowl, line 27: 💤

route-rs-runtime/src/link/primitive/process_link.rs
```diff
@@ -2,6 +2,10 @@ use crate::link::{Link, LinkBuilder, PacketStream, ProcessLinkBuilder};
 use crate::processor::Processor;
 use futures::{Async, Poll, Stream};
 
+/// `ProcessLink` is a simple links used to process packets through the user-defined processor
```
ddowl: "`ProcessLink` processes packets through a user-defined processor."

route-rs-runtime/src/link/primitive/process_link.rs
```diff
@@ -2,6 +2,10 @@ use crate::link::{Link, LinkBuilder, PacketStream, ProcessLinkBuilder};
...
+/// It does not have the ability to store packets internally, and is pull-based. So it only does
+/// work when it is called, and must either immediatly return recieved packets, or drop them.  This
+/// link is used for transformation logic.
```
ddowl: "It cannot buffer packets, so it only does work when it is called. It must immediately drop or return a transformed packet."

We can remove the last sentence.

route-rs-runtime/src/link/primitive/queue_link.rs
```diff
@@ -7,6 +7,8 @@ use crossbeam::crossbeam_channel::{Receiver, Sender, TryRecvError};
 use futures::{Async, Future, Poll, Stream};
 use std::sync::Arc;
 
+/// A link that can be used to create queues or buffers. It also takes a processor so
+/// packets may be transformed prior to being enqueued.
```
ddowl: Can remove "It also takes a processor so".

route-rs-runtime/src/link/utils/mod.rs
```diff
@@ -1,8 +1,8 @@
-/// Task Park is a structure for tasks to place their task handles when sleeping, and where they can
+/// Task Park is a structure for tasks to place their task handles in when sleeping, and where they can
 /// check for other tasks that need to be awoken.  As an example, the ingressor and egressor side of
```
ddowl: How about
> Task Park is a cache for task handles. Tasks can store their task handles there before sleeping, and can alert other tasks in the cache to be awoken. For example, ...

route-rs-runtime/src/pipeline/mod.rs
```diff
@@ -1,2 +1,8 @@
+//! # What are they?
+//!
+//! Pipelines are abstractions used by graphgen to define a grouping of links that are run by the runtime, in our case Tokio. The graphgen
```
ddowl: "grouping" sounds like just a set of links. Maybe we can use "graph" to imply that edges are established between them?

route-rs-runtime/src/pipeline/mod.rs
```diff
@@ -1,2 +1,8 @@
...
+//! program takes the user provided graph, and generates a pipeline. It connects input and output channels to the pipe, so that it can be
+//! hooked up to packet sources and sinks, and drops the pipeline into the runtime. Generally can be abstracted to the concept, router.
```
ddowl: > Generally can be abstracted to the concept, router.

Can you clarify this sentence? As in, the full Router program is composed of Pipelines?

route-rs-runtime/src/processor/mod.rs
```diff
@@ -1,3 +1,12 @@
+//! # What are they for?
+//!
+//! Processors are the unit of transformation in route-rs. The Processor is defined by a trait that requires that
+//! creators of processors implement a process function, that the underlying Link will call on every packet that
+//! moves through it. You can use processors to modify headers, count packets, update state, create new packets,
+//! join packets, whatever you want! As long as an processor conforms to the trait specification, the processor can be run inside a route-rs.
```
ddowl: > conforms to the trait specification,

"conforms to the `Processor` trait,"

route-rs-runtime/src/processor/mod.rs
```diff
@@ -10,6 +19,7 @@ pub use self::drop::*;
 mod dec_ip_hop;
 pub use self::dec_ip_hop::*;
 
+/// In order for a struct to become a Processor, it must implement this Trait.
```
ddowl: We can remove this comment. It doesn't seem to add any extra info for a user. e.g.
> In order for a struct to become Eq, it must implement Eq.

route-rs-runtime/src/utils/test/packet_collectors.rs
```diff
@@ -3,6 +3,8 @@ use crossbeam::crossbeam_channel::Sender;
 use futures::{Async, Future, Poll};
 use std::fmt::Debug;
 
+/// A structure that may be handed an input stream that it will exhaustively drain from until it
+/// recieves a None. Useful for testing purposes.
```
ddowl, line 7: Can remove "Useful for testing purposes". It exists in `utils/test` so I users can pick up on the context.

route-rs-runtime/src/link/mod.rs
CollinValley: Hmm, I guess if we control Tokio? We haven't quite nailed down the interface yet. I agree though this is confusing, I will remove it. I can bury this somewhere inside the Primitive page instead talking about how users should not need to define their own, since graphgen will not know about then among other things.

route-rs-runtime/src/pipeline/mod.rs
CollinValley: I'm not really sure what to do about pipelines right now. Ideally we won;t need them once we finish the router. The Link abstraction feels powerful enough, it can hook right to Afpacket or a PCAP input link.

route-rs-runtime/src/processor/mod.rs
CollinValley: lol that is brutally true

ddowl (approved):
LGTM!

PR #190, opened 2019-11-24
Packets builder improvement

CollinValley:
This is a series of commits that adjusts the interface for packets to be more ergonomic. Primarily, users now have the option of making packets either from a buffer, or as empty shells which they then populate.
To support this I also rounded out a lot of the methods on each class, so that relevant setters were available where it made sense. 
Finally, I added the ability to build a lower layer packet out of an upper layer one, by providing the encap function.

### DIFF
--- a/route-rs-packets/src/ipv4.rs
+++ b/route-rs-packets/src/ipv4.rs
@@ -51,6 +51,15 @@ impl Ipv4Packet {
         })
     }
 
+    /// Create an empty Ipv4Packet with no layer 2 header. All possible values are set to 0
+    pub fn empty() -> Ipv4Packet {
+        let mut data = vec![0x45];
+        data.resize(20, 0);
+        let total_len: u16 = 20;
+        data[2..=3].copy_from_slice(&total_len.to_be_bytes());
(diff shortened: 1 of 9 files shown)


route-rs-packets/src/ipv4.rs
```diff
@@ -153,10 +168,22 @@ impl Ipv4Packet {
         self.data[self.layer3_offset + 1] >> 2
     }
 
+    /// Lower 6 bits define the dcsp
```
eeverett6: nit: dscp

eeverett6: Also in the parameter

eeverett6 (commented):
I like this a lot, especially being able to easily encapsulate a layer 4 segment into a layer 3 packet, etc.

route-rs-packets/src/ipv4.rs
CollinValley: did I type that error again....ugh!

ddowl (approved):
Definitely seems reasonable to me. Thanks for adding extra tests for the new functions.

PR #195, opened 2020-01-12
Tokio bump

CollinValley:
This is a huge change, it needs some refactoring, particularly in the harness area.
Will probably layer on/rework, take a look at the harness code in particular, I think that is the most ugly of all of the reasons. But yeah, the bump actually worked!

Drew in particular the harness/test code could use some magic. The biggest gotcha right now is the PacketStreams, it they use a timer, need to be built in a function with the macro Tokio::main over it. That isn't amazing and maybe we can figure out how to work around that.


CollinValley:
Wrote in the test harness changes, and standardized all of them. It was quite a lot of work changing the setup and teardown of every test. But they all pass now, and we even solved a multithreaded runtime bug along the way

Next I need to resolve the rest cases (2) that are failing in Graphgen right now. If the travis checks pass something isn't working.  To do that I need to sit down for awhile with graphgen, which is something I've needed to do anyways. 

I'll try to have the graphgen tests fixed in the next couple days and then we will be ready for merge pending CR!

examples/dns-interceptor/src/pipeline.rs
```diff
@@ -67,11 +67,16 @@ impl route_rs_runtime::pipeline::Runner for Pipeline {
             .build_link();
         all_runnables.append(&mut runnables_6);
 
-        tokio::run(lazy(move || {
-            for r in all_runnables {
-                tokio::spawn(r);
+        let mut rt = runtime::Builder::new().enable_all().build().unwrap();
+
+        rt.block_on(async move {
+            let mut handles = vec![];
```
ddowl: Construction via `map/fold`? _#unnecessarilyfunctional_

route-rs-runtime/src/link/composite/drop_link.rs
```diff
@@ -119,6 +119,11 @@ mod tests {
 
     #[test]
     fn finishes_with_wait() {
+        run_finishes_with_wait();
+    }
+
+    #[tokio::main]
```
ddowl: Can we annotate `finishes_with_wait` with multiple macros so we don't need to dispatch to an extra function?

route-rs-runtime/src/link/mod.rs
```diff
@@ -28,9 +27,9 @@ pub mod primitive;
 pub mod utils;
 
 /// All Links communicate through streams of packets. This allows them to be composable.
-pub type PacketStream<Input> = Box<dyn Stream<Item = Input, Error = ()> + Send>;
+pub type PacketStream<Input> = Box<dyn futures::Stream<Item = Input> + Send + Unpin>;
```
ddowl, line 30: Can you describe why `PacketStream`s are `Unpin`? Is this just explicitly stating that packets may be moved/copied across threads/cores? I thought [most types implement `Unpin` by default](https://doc.rust-lang.org/std/marker/trait.Unpin.html), but a generic type needs to be marked?

ddowl: Maybe without `Unpin` we would need to enforce that every packet is `Pin`ned, and that's not a restriction we want to make? https://rust-lang.github.io/async-book/04_pinning/01_chapter.html#how-to-use-pinning

route-rs-runtime/src/link/primitive/classify_link.rs
```diff
@@ -163,6 +166,8 @@ pub struct ClassifyIngressor<'a, C: Classifier> {
     task_parks: Vec<Arc<AtomicCell<TaskParkState>>>,
 }
 
+impl<'a, C: Classifier> Unpin for ClassifyIngressor<'a, C> {}
```
ddowl, line 169: Can we `#[derive(Unpin)]`?

route-rs-runtime/src/link/primitive/fork_link.rs
```diff
@@ -149,22 +151,21 @@ impl<P> Drop for ForkIngressor<P> {
 }
 
 impl<P: Send + Clone> Future for ForkIngressor<P> {
-    type Item = ();
-    type Error = ();
+    type Output = ();
 
     /// If any of the channels are full, we await that channel to clear before processing a new packet.
-    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
```
ddowl, line 143: Why declare `PacketStream`s `Unpin` if packets are `Pin`?

ddowl: I'm not asking to be snarky, I genuinely don't know.

route-rs-runtime/src/link/primitive/input_channel_link.rs
```diff
@@ -48,15 +50,16 @@ struct StreamFromChannel<Packet> {
     channel_receiver: crossbeam::Receiver<Packet>,
 }
 
+impl<Packet> Unpin for StreamFromChannel<Packet> {}
```
ddowl, line 53: Can we `#[derive(Unpin)]`?

route-rs-runtime/src/link/primitive/join_link.rs
```diff
@@ -109,6 +111,8 @@ pub struct JoinIngressor<Packet: Sized> {
     task_park: Arc<AtomicCell<TaskParkState>>,
 }
 
+impl<Packet: Sized> Unpin for JoinIngressor<Packet> {}
```
ddowl, line 114: Can we `#[derive(Unpin)]`?

route-rs-runtime/src/link/primitive/join_link.rs
```diff
@@ -208,35 +216,37 @@ impl<Packet: Sized> Drop for JoinEgressor<Packet> {
 
 impl<Packet: Sized> Stream for JoinEgressor<Packet> {
     type Item = Packet;
-    type Error = ();
 
     /// Iterate over all the channels, pull the first packet that is available.
     /// This starts at the next index after the last successful recv
-    fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Option<Self::Item>> {
```
ddowl, line 214: Comment: I like that `Future`s and `Stream`s have different poll fns `poll / poll_next` to help distinguish them.

route-rs-runtime/src/link/primitive/input_channel_link.rs
```diff
@@ -48,15 +50,16 @@ struct StreamFromChannel<Packet> {
...
+
 impl<Packet> Stream for StreamFromChannel<Packet> {
     type Item = Packet;
-    type Error = ();
 
-    fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
+    fn poll_next(self: Pin<&mut Self>, _cx: &mut Context) -> Poll<Option<Self::Item>> {
```
ddowl, line 58: Not sure why `poll_next` must take `Pin<&mut Self>` as the first argument, since the [documentation](https://docs.rs/futures/0.2.0/futures/stream/trait.Stream.html#tymethod.poll_next) says the type bound is just `&mut Self`. However, `poll` [_is_ restricted](https://doc.rust-lang.org/std/future/trait.Future.html#tymethod.poll) to `Pin<&mut Self>`.

route-rs-runtime/src/link/primitive/join_link.rs
```diff
@@ -348,8 +358,18 @@ mod tests {
         assert_eq!(results[0].len(), stream_len * num_streams);
     }
 
+    // This test calls out to a helper async function because async fn are not allowed in tests yet. This allows us to use
+    // the provided macro tokio::main, which starts up a runtime. The runtime needs to be created first before we can make a
+    // packet_generator, since the packet generator needs to create an Interval, which needs to hook into a pre-created timer
+    // thread running in Tokio. As such, we also have a different harness function, execute_link, which is async. Perhaps this
+    // would be a better
```
ddowl: I'd be happy to look into this later.

route-rs-runtime/src/link/primitive/join_link.rs
```diff
@@ -348,8 +358,18 @@ mod tests {
...
     #[test]
     fn wait_between_packets() {
+        run_wait_between_packets();
+    }
+
+    #[tokio::main]
```
ddowl: Same as above, can we annotate a function with multiple macros?

route-rs-runtime/src/link/primitive/process_link.rs
```diff
@@ -88,38 +90,37 @@ impl<P: Processor> ProcessRunner<P> {
...
 
+impl<P: Processor> Unpin for ProcessRunner<P> {}
+
 impl<P: Processor> Stream for ProcessRunner<P> {
     type Item = P::Output;
-    type Error = ();
 
     /// Intro to `Stream`s:
-    /// 4 cases: `Async::Ready(Some)`, `Async::Ready(None)`, `Async::NotReady`, `Err`
+    /// 3 cases: `Poll::Ready(Some)`, `Poll::Ready(None)`, `Poll::Pending`
```
ddowl, line 99: Lol see ya `Err` 👋

route-rs-runtime/src/link/utils/task_park.rs
```diff
@@ -82,10 +82,9 @@ pub fn unpark_and_notify(task_park: &Arc<AtomicCell<TaskParkState>>) {
 /// Notifies a task if it resides in the `task_park`, and
 /// then parks the callee task in the `task_park`.
 /// Use when you wish to sleep the current task
-pub fn park_and_notify(task_park: &Arc<AtomicCell<TaskParkState>>) {
-    let task = task::current();
-    if !swap_and_notify(task_park, TaskParkState::Parked(task)) {
-        task::current().notify();
+pub fn park_and_notify(task_park: &Arc<AtomicCell<TaskParkState>>, task: task::Waker) {
```
ddowl: Maybe rename to `park_and_wake` to fit new API?

route-rs-runtime/src/utils/test/harness.rs
```diff
@@ -46,8 +46,8 @@ pub fn run_link<OutputPacket: Debug + Send + Clone + 'static>(
     // gather link's runnables and tokio-driven consumers into one collection
     runnables.append(&mut consumers);
 
-    // 🏃💨💨
```
ddowl, line 49: 😞

route-rs-runtime/src/utils/test/harness.rs
```diff
@@ -56,11 +56,60 @@ pub fn run_link<OutputPacket: Debug + Send + Clone + 'static>(
...
+pub fn initialize_runtime() -> runtime::Runtime {
+    runtime::Builder::new()
+        .threaded_scheduler()
+        .enable_all()
+        .build()
+        .unwrap()
+}
+
+// This is kind of ugly ish, right now
+pub async fn execute_link<OutputPacket: Debug + Send + Clone + 'static>(
```
ddowl: Personally I'd prefer modifying `run_link` rather than leaving an old copy around. We can revisit the ugliness later if we need to.

route-rs-runtime/src/utils/test/harness.rs
```diff
@@ -76,8 +31,7 @@ pub fn initialize_runtime() -> runtime::Runtime {
         .unwrap()
 }
 
-// This is kind of ugly ish, right now
-pub async fn execute_link<OutputPacket: Debug + Send + Clone + 'static>(
+pub async fn run_link<OutputPacket: Debug + Send + Clone + 'static>(
```
ddowl, line 34: Initializing tests in the async runtime context is fine for now. It might be interesting to see if we could get a `#[with_tokio_runtime]` macro working so we could annotate `async fn`s. That would be a nice easy win to reduce boilerplate and 1 level of indentation.

route-rs-runtime/src/link/primitive/classify_link.rs
```diff
@@ -221,7 +208,17 @@ impl<'a, C: Classifier> Future for ClassifyIngressor<'a, C> {
                 ready!(Pin::new(&mut ingressor.input_stream).poll_next(cx));
 
             match packet_option {
-                None => return Poll::Ready(()),
+                None => {
```
ddowl, line 211: It seems like each link implements the "eggressor check" in a slightly different way. Is this due to actual different contexts among links, or are they common enough that we can abstract them into the `task_park` module?

examples/minimal-static-router/src/classifiers.rs
```diff
@@ -121,31 +121,32 @@ mod tests {
...
+        ];
+        let data_unknown: Vec<u8> = vec![
+            0xde, 0xad, 0xbe, 0xef, 0xff, 0xff, 1, 2, 3, 4, 5, 6, 0xff, 0xff,
+        ];
+        let frame1 = EthernetFrame::from_buffer(data_v4, 0).unwrap();
+        let frame2 = EthernetFrame::from_buffer(data_v6, 0).unwrap();
+        let frame3 = EthernetFrame::from_buffer(data_unknown, 0).unwrap();
+
+        let mut runtime = initialize_runtime();
+        let results = runtime.block_on(async {
```
ddowl, line 147: We may want to encapsulate this pattern into the test harness later

ddowl (changes requested):
Overall, seems like a pretty straightforward mechanical change with a few unforeseen minor bugs. We can either address my comments or tag them for later work so we can push the version bump through. Good job @CollinValley !

route-rs-runtime/src/link/composite/drop_link.rs
CollinValley: This is changed in the next commit where I rework the testing harness.  I just reworked run_link and such. See the test harness commit. I think that resolves this.

ddowl: Oh yes I believe you're right. I was leaving comments as I read through each commit (they're a little more bite-sized, focused), so we can resolve them if you already addressed the comment.

route-rs-runtime/src/link/mod.rs
CollinValley: This isn't that the packets need to be Unpin, it is the object that provides the Stream must be unpin. Unpin means we can freely pin and then unpin the object. Structs wont implement unpin by default because they may contain pointers and such that could move. They could provide a custom implementation if that were the case, but since the ingressor and egressor own all their components, it isn't an issue to just add the trait to get the included 'provided' methods. It is because when we call poll_next on that stream object, we need to pin it in place so the futures work correctly.  AFAIK, this is because the runtime needs to know that the object doesn't move out from underneath it when it is trying to reference the future. The poll_next function can't move on the runtime otherwise its calls would become invalidated it the function pointer they stored no longer pointed at the poll function cause the object moved in memory.  Something like that, I think the withoutboats talk may be more illuminating.  But in essence, we need our objects that implement packetstream to impl unpin so that I can shove them into a Pin object to call their poll_next() function.  Checkout the function sig of poll_next to see why.

CollinValley: I think unpin is kind of a bad name, but I'm not sure if I can think of anything better.

route-rs-runtime/src/link/primitive/classify_link.rs
CollinValley: Hmm...maybe.

CollinValley: Looks like no, that macro doesn't exist...yet?

route-rs-runtime/src/link/mod.rs
CollinValley: Okay I think this will be informative: Take a look at ClassifyIngessor's Future impl
`impl<'a, C: Classifier> Future for ClassifyIngressor<'a, C> {
    type Output = ();

    /// Same logic as QueueEgressor, except if any of the channels are full we
    /// await that channel to clear before processing a new packet. This is somewhat
    /// inefficient, but seems acceptable for now since we want to yield compute to
    /// that egressor, as there is a backup in its queue.
    fn poll(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
        let ingressor = Pin::into_inner(self);
        loop {
            for (port, to_egressor) in ingressor.to_egressors.iter().enumerate() {
                if to_egressor.is_full() {
                    park_and_notify(&ingressor.task_parks[port], cx.waker().clone());
                    return Poll::Pending;
                }
            }

            //TODO: Standardize in_stream, input_stream, and stream to one name
            let packet_option: Option<C::Packet> =
                ready!(Pin::new(&mut ingressor.input_stream).poll_next(cx));

            match packet_option {
                None => {
                    for to_egressor in ingressor.to_egressors.iter() {
                        to_egressor
                            .try_send(None)
                            .expect("ClassifyIngressor::Drop: try_send to_egressor shouldn't fail");
                    }
                    for task_park in ingressor.task_parks.iter() {
                        die_and_notify(&task_park);
                    }
                    return Poll::Ready(());
                }
                Some(packet) => {
                    let class = ingressor.classifier.classify(&packet);
                    let port = (ingressor.dispatcher)(class);
                    if port >= ingressor.to_egressors.len() {
                        panic!("Tried to access invalid port: {}", port);
                    }
                    if let Err(err) = ingressor.to_egressors[port].try_send(Some(packet)) {
                        panic!(
                            "Error in to_egressors[{}] sender, have nowhere to put packet: {:?}",
                            port, err
                        );
                    }
                    unpark_and_notify(&ingressor.task_parks[port]);
                }
            }
        }
    }
}`

The first interesting line: `        let ingressor = Pin::into_inner(self);`
This ingressor, when the runtime calls it via Poll(..), has a Pinned version of the ingressor for self.  So to actually get at the internal values, we need to unpin it, so then we can refer to our methods and member variables as usual. If this were not an unpin struct, ie it was self referential, it wouldn't work. If it is self-referential or something like that, once you Pin it in place, you say "This can't be moved without invalidating something", which in the this example would be the internal pointer to itself.  Prior to Pin I don't think being self referential was something the rust lang supported. 

The second line: `ready!(Pin::new(&mut ingressor.input_stream).poll_next(cx));` Is the same thing but in reverse. Before we can call poll_next on our Stream object, we need to tell the computer to pin the Stream in place, so to satisfy the requirements of poll_next and poll, where self is `self: Pin<&mut Self>`

route-rs-runtime/src/link/primitive/fork_link.rs
CollinValley: Packets have no Pin requirement. In your example, ForkIngressor the object must be Pinned before one may call poll on it. 
For example

```
let cx = Context; //We got this variable from somewhere, not important for this example.  
let forkingressor = ForkIngressor::new() //lets pretend this works
forkingressor.poll(cx); //Compiler error: method not found
Pin::new(&mut forkingressor).poll(cx);  //Happy rustc
```

route-rs-runtime/src/link/primitive/input_channel_link.rs
CollinValley: :bigsad: no

route-rs-runtime/src/link/primitive/input_channel_link.rs
CollinValley: You've been smekledorfed by old documentation https://docs.rs/futures/0.3.1/futures/stream/trait.Stream.html

route-rs-runtime/src/link/primitive/join_link.rs
CollinValley: Nah

route-rs-runtime/src/link/utils/task_park.rs
CollinValley: not a bad idea. I'll try to think about some names that will work.

route-rs-runtime/src/utils/test/harness.rs
CollinValley: Oh, I can add the emoji back in, in later commits.

route-rs-runtime/src/utils/test/harness.rs
CollinValley: Yeah, in the later commit, I did just that, execute link was renamed to run_link in a couple commits.

route-rs-runtime/src/link/primitive/classify_link.rs
CollinValley: Hmm, they are the same in the sense that you iterate over the entire collection of egressors, and then die_and_notify for each associated task park. So in this case, fork_link and classify_link have the same flow, and then join_link is its own case, and queue_link is the base case. So, I'm not sure there is much of a common piece we can cut out, without taking additional args into the function. 

I think I'm going to leave the the logic as is, while it isn't the most DRY, it is pretty readable right now.

route-rs-runtime/src/utils/test/harness.rs
CollinValley: Yeah, macros are kind of scary but yeah we should try to figure out some sort of macro. Will think about this for a future improvement.

examples/minimal-static-router/src/classifiers.rs
CollinValley: Agree

route-rs-runtime/src/utils/test/harness.rs
CollinValley: Small issue is that the test fn's can not be async. Which is annoying, but we may be able to work around.

route-rs-runtime/src/link/utils/task_park.rs
CollinValley: subbing wake should work pretty well, do we still like park though?

route-rs-runtime/src/link/primitive/join_link.rs
CollinValley: This ended up being closed by the harness rework. The harness could probably use some improvement via macro, but the initialize_runtime then runtime.block_on(... is likely what we are going to go with for the time being.

route-rs-runtime/src/link/mod.rs
ddowl: > Structs wont implement unpin by default because they may contain pointers and such that could move. They could provide a custom implementation if that were the case, but since the ingressor and egressor own all their components, it isn't an issue to just add the trait to get the included 'provided' methods.

It feels like the compiler should be able to do some kind of analysis on structs to determine whether or not this is the case, but i don't wanna yell at the core team. I'm just happy `async` is stable.

That example was pretty informative. I'm going to watch withoutboats' talk on `Pin` to refresh my knowledge, but I'm going to trust your judgement 👌

route-rs-runtime/src/link/utils/task_park.rs
ddowl: I still like "park".

route-rs-runtime/src/utils/test/harness.rs
ddowl: 💯

route-rs-runtime/src/utils/test/harness.rs
ddowl: That's so annoying! There will always be more "make testing better" work todo.

examples/dns-interceptor/src/pipeline.rs
ddowl: :heck:

route-rs-runtime/src/utils/runner.rs
```diff
@@ -0,0 +1,28 @@
+use tokio::runtime;
+use crate::link::{TokioRunnable};
+use tokio::task::JoinHandle;
+
+/// Runner is a user facing helper function for running the constructed router.TokioRunnable
```
ddowl: >  for running the constructed router.TokioRunnable

typo?

route-rs-runtime/src/utils/runner.rs
```diff
@@ -21,6 +21,7 @@ pub fn runner(link_builder: fn() -> Vec<TokioRunnable>) {
     runtime.block_on(async {
         let runnables = link_builder();
         let handles: Vec<JoinHandle<()>> = runnables.into_iter().map(|runnable| tokio::spawn(runnable)).collect();
+        // 🏃💨💨
```
ddowl, line 50: 🥇 yeet

route-rs-runtime/src/utils/runner.rs
CollinValley: yee, wil fixup

ddowl (dismissed):
Looks like this branch needs another dose of `clippy` before Travis accepts it. Otherwise, LGGTMMD. Congrats on merging stable `async`!

CollinValley:
I'll need to fixup graphgen tests before I am ready to fully merge, but it is close

route-rs-runtime/src/link/utils/task_park.rs
CollinValley: named park_and_wake

examples/dns-interceptor/src/pipeline.rs
CollinValley: gong to stick with simple for now here, when we revamp the new public facing running harness then I for one welcome our functional overlords.

CollinValley:
This all looks good to me, will wait for Drew to do the honors

ddowl (approved):
gg Sam. Homie Stamp

PR #196, opened 2020-01-16
Codegen improvements

scgruber:
This PR is primarily about adding some helpers to condense the code in gen_tokio_run. Some of these helpers could be used more widely but we are holding off on that during the work to convert to tokio 0.2.


CollinValley (approved):
Looks good to me, really cuts down on the code in some places. Good refactor.

PR #198, opened 2020-02-02
Codegen cleanup

scgruber:
This is a bunch of small commits that clean up our code generation. Some of it is just using the existing codegen functions in places where the raw `syn` objects are generated in `main.rs`, and some of it is writing new helpers. At the end, we have gotten rid of all of the instances of `proc_macro2::Span::call_site()` in `main.rs`, which feels much better.

Fair warning, theres a lot of cargo fmt churn in this one, so looking at the individual commits will be helpful to understand what's actually changing.

### DIFF
--- a/route-rs-graphgen/src/codegen.rs
+++ b/route-rs-graphgen/src/codegen.rs
@@ -330,21 +330,6 @@ mod typedef {
     }
 }
 
-pub fn angle_bracketed_types(types: Vec<syn::Type>) -> syn::AngleBracketedGenericArguments {
-    syn::AngleBracketedGenericArguments {
-        colon2_token: None,
-        lt_token: syn::token::Lt {
-            spans: [fake_span()],
-        },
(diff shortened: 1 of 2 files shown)


route-rs-graphgen/src/codegen.rs
```diff
@@ -742,6 +742,16 @@ pub fn expr_match(input: syn::Expr, arms: Vec<(syn::Pat, syn::Expr)>) -> syn::Ex
     })
 }
 
+pub fn expr_lit_int<N>(number: N) -> syn::Expr
+where
+    N: std::string::ToString,
```
CollinValley, line 747: Yeah, I assume this is a workaround without const generics, which are being added to help fix this issue.

CollinValley (approved):
Awesome work!  These changes are really making the codegen code much more readable, should make maintenance easier too. 
LGTMMD

PR #199, opened 2020-02-06
MVR: Skeleton for new router structure

scgruber:
This commit defines the basic structure for making a router using the
new framework with tokio 0.2.

In main() we create:
  - ingressors
  - egressors
  - tokio runtime
  - Route instance

We then pass these to run_route() to wire everything together and
initiate the tokio threads. run_route() should be moved into
route-rs-runtime before all of this is actually merged, since it's
pretty generic over any route link.

Since we don't have af_packet integrated with links yet, we're still
setting up InputChannelLink and OutputChannelLink at either end of
the router. This new framework does handle multiple inputs and output
channels, though.

While this does compile, it probably won't run because a lot of things
are left empty. I'll fill them in with reasonable defaults before this
is merged. A following commit will implement the top-level internals
of the minimum viable router, and then a further commit will convert
graphgen and the other examples into this style.

### DIFF
--- a/examples/minimum-viable-router/src/main.rs
+++ b/examples/minimum-viable-router/src/main.rs
@@ -3,6 +3,7 @@ use route_rs_runtime::link::primitive::{InputChannelLink, OutputChannelLink};
 use route_rs_runtime::link::{LinkBuilder, PacketStream, TokioRunnable};
 use tokio::task::JoinHandle;
 
+mod classifier;
 mod processor;
 mod route;
 
--- a/examples/minimum-viable-router/src/main.rs
+++ b/examples/minimum-viable-router/src/main.rs
@@ -3,6 +3,7 @@ use route_rs_runtime::link::primitive::{InputChannelLink, OutputChannelLink};
 use route_rs_runtime::link::{LinkBuilder, PacketStream, TokioRunnable};
 use tokio::task::JoinHandle;
 
+mod processor;
 mod route;
 
 fn main() {
--- a/examples/minimum-viable-router/src/main.rs
+++ b/examples/minimum-viable-router/src/main.rs
@@ -1,0 +1,85 @@
+use route_rs_packets::EthernetFrame;
+use route_rs_runtime::link::primitive::{InputChannelLink, OutputChannelLink};
+use route_rs_runtime::link::{LinkBuilder, PacketStream, TokioRunnable};
+use tokio::task::JoinHandle;
+
+mod route;
+
+fn main() {
+    let mut runtime = tokio::runtime::Builder::new()
+        .threaded_scheduler()
+        .enable_all()
+        .build()
+        .unwrap();
+
+    let ingress_wan = crossbeam::channel::unbounded();
+    let ingress_lan = crossbeam::channel::unbounded();
+    let ingress_host = crossbeam::channel::unbounded();
+
+    let egress_wan = crossbeam::channel::unbounded();
+    let egress_lan = crossbeam::channel::unbounded();
+    let egress_host = crossbeam::channel::unbounded();
+
+    drop(ingress_wan.0);
+    drop(ingress_lan.0);
+    drop(ingress_host.0);
+
+    run_route(
+        vec![ingress_wan.1, ingress_lan.1, ingress_host.1],
+        vec![egress_wan.0, egress_lan.0, egress_host.0],
+        &mut runtime,
+        route::Route::new(),
+    );
+
+    assert_eq!(egress_wan.1.len(), 0, "Egress WAN is not empty");
+    assert_eq!(egress_lan.1.len(), 0, "Egress LAN is not empty");
+    assert_eq!(egress_host.1.len(), 0, "Egress Host is not empty");
+
+    println!("Congratulations! Your router successfully does nothing!")
+}
+
+// TODO: Move this to route-rs-runtime
+fn run_route(
+    ingressors: Vec<crossbeam::Receiver<EthernetFrame>>,
+    egressors: Vec<crossbeam::Sender<EthernetFrame>>,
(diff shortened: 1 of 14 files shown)


examples/minimum-viable-router/src/main.rs
```diff
@@ -0,0 +1,62 @@
...
+            .map(|(ingressor, sender)| {
+                OutputChannelLink::new()
+                    .ingressor(ingressor)
+                    .channel(sender)
+                    .build_link()
+            })
+            .unzip();
+
+    all_runnables.append(&mut egress_runnables.into_iter().flatten().collect());
+
```
CollinValley, line 79: All this building needs to happen inside the block_on block, otherwise you can get initialization failures.  The block on provides the runtime context. It will be a rare error, but if people have external timer links and stuff it won't run.

CollinValley: Especially because if Nathans links are initialized outside of the "router", those will need the tokio context active since they need to access the reactor and executor threads of tokio.

scgruber: The way I envisioned this is that the `ingressors` and `egressors` would be replaced with the socket objects and we'd still assemble the links inside of `run_route`. Probably given that we should have a less generic name for this method, since it bakes in the use of `af_packet`.

But it should be relatively okay to just extend the `block_on` around the rest of this method body.

CollinValley: yeah, that makes sense to me. The block_on thing is annoying,  Another annoyance is that the `route` object needs to be also built inside a tokio context.  So we should probably change that argument to something like `route: Fn() -> Link`, then we can call `let router = route::new();` inside the tokio `block_on`.  For an example of this, I did it inside the prototype run harness in the runtime, under `utils`

CollinValley: The Fn doesn't have to return a `Link` either, could be some other struct, but returning the `Link` makes some sense from our composability story.

scgruber: Do we have to initialize the `route` object fully inside the `block_on`? Or can we just `build_link()` in there. Keeping it outside allows the implementor to pass additional setup fields in `main()` if desired.

CollinValley: Only the "Link" objects need to be initialized inside the block_on, so yes we could pre-init all the builder syntax stuff outside the context, then pass the object that you are about to build_link() on into the context, and call build_link.   I think that is a pretty good idea.

examples/minimum-viable-router/src/main.rs
```diff
@@ -0,0 +1,85 @@
...
+    let ingress_wan = crossbeam::channel::unbounded();
+    let ingress_lan = crossbeam::channel::unbounded();
+    let ingress_host = crossbeam::channel::unbounded();
+
+    let egress_wan = crossbeam::channel::unbounded();
+    let egress_lan = crossbeam::channel::unbounded();
+    let egress_host = crossbeam::channel::unbounded();
+
+    drop(ingress_wan.0);
```
CollinValley, line 25: Why are we doing this? We can just let Rust do the dropping for us? Also won't the recv close as soon as we do this?

route-rs-runtime/src/link/primitive/black_hole_link.rs
```diff
@@ -0,0 +1,91 @@
+use crate::link::{Link, LinkBuilder, PacketStream};
```
ddowl, line 1: I thought we got rid of the black hole links in favor of moving that logic to Processors?

examples/minimum-viable-router/src/main.rs
```diff
@@ -0,0 +1,85 @@
...
+    );
+
+    assert_eq!(egress_wan.1.len(), 0, "Egress WAN is not empty");
+    assert_eq!(egress_lan.1.len(), 0, "Egress LAN is not empty");
+    assert_eq!(egress_host.1.len(), 0, "Egress Host is not empty");
+
+    println!("Congratulations! Your router successfully does nothing!")
+}
+
+// TODO: Move this to route-rs-runtime
```
CollinValley, line 43: This looks exactly like what already exists in the runtime under the utils called "runner"

examples/minimum-viable-router/src/main.rs
scgruber: That is the point. It looks a little silly here, but if we actually had packets to pass through, this would allow us to run a test in the `main()`. If we don't explicitly drop the channels, `run_route()` will never exit because nothing stops the router.

route-rs-runtime/src/link/primitive/black_hole_link.rs
scgruber: There's no way to poll the links, so we'd have to do something really strange like join the drop path back into one of the actual flows of the router.

examples/minimum-viable-router/src/main.rs
scgruber: `runner` doesn't provide a way to configure the inputs to the overall `Link`. I suppose you could put it all into the thunk that returns the `Link`, but I'm trying to keep that logic inside the library.

Once we have `af_packet`, this method will handle setting up all of those links and take the socket paths instead of channels.

scgruber: May be able to refactor this thing to call `runner` though.

CollinValley: Hmm, I just want runner to be the thing that we use.  I'm not married to the current implementation of runner per se, just that this commit stack should make runner fit our purpose here. Let me think on it for a few days and find a way to merge everything. This use case is valid.

PR #200, opened 2020-02-11
Interface tagging work

CollinValley:
Start of my interface tagging work.  I've opted to do a bit of a hybrid solution, we have a new wrapper struct (src/types.rs) that provides inbound and outbound interface tags.  This first link, called interface_mux (not hot on the name, inbound_mux, inbound_join? input_*)? takes in our 3 known interfaces and tags them with the `Interface` enum.  Note the wrapper is generic so we can up/down the inner packet, without needing to redo the definition, but our impl is not generic, and contains only defined types.


CollinValley:
Check out the test to get an idea of how it works now.

PR #201, opened 2020-02-12
afpacket integration with route-rs

moosingin3space:


route-rs-afpacket/tests/integration_test.rs
```diff
@@ -0,0 +1,168 @@
...
+
+#[test]
+#[should_panic]
+fn input_panics_when_built_without_channel() {
+    AfPacketInputLink::new().build_link();
+}
+
+#[test]
+#[ignore]
+fn input_identity() {
```
CollinValley, line 102: Why are these tests ignored? If they can't be run by the CI can you leave a small comment about how to run them locally?

CollinValley:
Rather than create a whole new crate, let's bring these links into the runtime itself.  They are Links in their own right, and then the runtime can just be dependent on a version of afpacket in the future.

Cargo.toml
```diff
@@ -8,6 +8,7 @@ members = [
 
     # I/O Crates
     "afpacket",
+    "route-rs-afpacket",
```
CollinValley, line 11: Nit: I'd like these links to be moved into the runtime, since all the primitive links reside there, and this is low level enough (Read: actually implements Future/Stream) that I think it has a good home there. Also I would like the runtime crate itself to ship with a complete set of tools, which having I/O links is a requirement of.

I think you should put this whole thing in `route-rs-runtime/links/primitive` and have a cfg-linux flag on the relevant mods.

I am leaving it up to you though, I'll accept wherever you want to stash it in the end, hence Nit.

afpacket/src/tokio_sockets.rs
```diff
@@ -20,11 +35,287 @@ impl AsyncBoundSocket {
         self.sock.get_mut().set_promiscuous(p)
     }
 
-    pub async fn send(&mut self, frame: &[u8]) -> io::Result<usize> {
-        self.sock.write(frame).await
+    pub fn split(self) -> (SendHalf, RecvHalf) {
```
CollinValley, line 38: Documentation: Something about the purpose of split, what it is used for.  Reference to Tokio maybe.

afpacket/src/tokio_sockets.rs
```diff
@@ -20,11 +35,287 @@ impl AsyncBoundSocket {
...
+
+        let rx = RecvHalf {
+            inner: inner.clone(),
+        };
+        let tx = SendHalf { inner };
+
+        (tx, rx)
+    }
+
+    pub fn poll_can_rx(&mut self, cx: &mut Context<'_>) -> Poll<io::Result<()>> {
```
CollinValley, line 52: Docs: When does this return Ready, when does it return Pending.  Add something for each of these poll_* functions. 
Should also specify that these functions are non-blocking.  Just think of all the things you want to know when you work with a library for functions like these.

afpacket/src/tokio_sockets.rs
CollinValley: You don't have to go crazy here, but some documentation is needed for sure. Will help with maintenance in the future as well. :)

afpacket/src/tokio_sockets.rs
```diff
@@ -20,11 +35,287 @@ impl AsyncBoundSocket {
...
+        ready!(self.sock.poll_read_ready(cx, ready))?;
+        Poll::Ready(Ok(()))
+    }
+
+    pub fn poll_can_tx(&mut self, cx: &mut Context<'_>) -> Poll<io::Result<()>> {
+        ready!(self.sock.poll_write_ready(cx))?;
+        Poll::Ready(Ok(()))
+    }
+
+    pub fn clear_can_tx(&mut self, cx: &mut Context<'_>) -> io::Result<()> {
```
CollinValley, line 63: Maybe explain why one needs to clear?

afpacket/src/tokio_sockets.rs
```diff
@@ -20,11 +35,287 @@ impl AsyncBoundSocket {
...
+                Poll::Pending
+            }
+            Err(e) => Poll::Ready(Err(e)),
+        }
+    }
+
+    pub fn poll_recv(
+        &mut self,
+        cx: &mut Context<'_>,
+        frame: &mut [u8],
```
CollinValley, line 87: Are there size requirements for frame?  What happens if the data to write would overflow the buffer? Panic? Should be documented.

afpacket/src/tokio_sockets.rs
```diff
@@ -20,11 +35,287 @@ impl AsyncBoundSocket {
...
+
+    pub fn recv<'a>(
+        &'a mut self,
+        frame: &'a mut [u8],
+    ) -> impl Future<Output = io::Result<(usize, sockets::Addr)>> + 'a {
+        RecvHalfRecvFuture { half: self, frame }
+    }
+}
+
+struct RecvHalfRecvFuture<'a> {
```
CollinValley, line 212: Let's get some doc comments on what all these structs are for. They won't be visible in the public facing documentation, but useful for maintenance.

afpacket/src/tokio_sockets.rs
```diff
@@ -20,11 +35,287 @@ impl AsyncBoundSocket {
...
+
+pub struct SendHalf {
+    inner: Arc<Inner>,
+}
+
+pub struct SendHalfLock<'a> {
+    guard: Guard<'a>,
+}
+
+impl RecvHalf {
```
CollinValley, line 171: Nit: I like Impl blocks near a struct definition if possible.

afpacket/tests/afpacket_tokio_tests.rs
```diff
@@ -65,3 +65,62 @@ fn layer2_loopback() {
...
+
+        tx_b.send(&eth_pkt.data).await.unwrap();
+
+        let mut in_buffer = vec![0; 1500];
+        let (len, _) = time::timeout(timeout, rx_a.recv(&mut in_buffer))
+            .await
+            .unwrap()
+            .unwrap();
+        in_buffer.resize(len, 0);
+        assert_eq!(in_buffer, eth_pkt.data);
```
CollinValley, line 122: Asserts need to be outside of the runtime context to cause a test to pass/fail.

route-rs-afpacket/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+license = "MIT"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+afpacket = { path = "../afpacket", features = ["tokio-support"] }
+route-rs-runtime = { path = "../route-rs-runtime" }
+futures = "0.3.4"
+
+[dev-dependencies]
```
CollinValley, line 15: Nice touch here with dev-dependencies for test only imports

route-rs-afpacket/src/input.rs
```diff
@@ -0,0 +1,70 @@
...
+    self, ready,
+    task::{Context, Poll},
+};
+use route_rs_runtime::link;
+use std::{mem, pin::Pin};
+
+const MTU: usize = 1500;
+
+#[derive(Default)]
+pub struct AfPacketInputLink {
```
CollinValley, line 12: Documentation needed for this link.  Especially about how to use it on linux systems, etc.

route-rs-afpacket/src/input.rs
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    pub fn channel(self, ingress: afpacket::RecvHalf) -> Self {
+        Self {
+            ingress: Some(ingress),
+        }
+    }
+}
+
+impl link::LinkBuilder<(), Vec<u8>> for AfPacketInputLink {
```
CollinValley, line 28: Is it possible for AfPacketInputLink to return half a packet in one recv, then the other half on the next recv? Or does it only return discreet packets?

route-rs-afpacket/src/input.rs
```diff
@@ -0,0 +1,70 @@
...
+impl Unpin for Stream {}
+
+impl futures::Stream for Stream {
+    type Item = Vec<u8>;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let me = self.get_mut();
+        let (sz, _) =
+            ready!(me.ingress.poll_recv(cx, &mut me.in_buf)).expect("failed to read packet");
+        me.in_buf.resize(sz, 0);
```
CollinValley, line 66: Why do we resize this vector on each received size? Why not just leave it at MTU length always?

route-rs-afpacket/src/output.rs
```diff
@@ -0,0 +1,100 @@
...
+use futures::{
+    self, ready,
+    task::{Context, Poll},
+    Future, Stream,
+};
+use route_rs_runtime::link;
+use std::pin::Pin;
+
+#[derive(Default)]
+pub struct AfPacketOutputLink {
```
CollinValley, line 11: Docs for the link.

route-rs-afpacket/src/output.rs
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        Self {
+            in_stream: Some(in_stream),
+            egress: self.egress,
+        }
+    }
+
+    fn build_link(self) -> link::Link<()> {
+        match (self.in_stream, self.egress) {
```
CollinValley, line 62: Nifty way of checking those values are not none, nice.

route-rs-afpacket/src/output.rs
```diff
@@ -0,0 +1,100 @@
...
+    stream: link::PacketStream<Vec<u8>>,
+    egress: afpacket::SendHalf,
+}
+
+impl Unpin for StreamToChannel {}
+
+impl Future for StreamToChannel {
+    type Output = ();
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
```
CollinValley, line 86: Could you doc comment here the cases where it fails to write, how it gets to lock, how it retries, etc?

route-rs-afpacket/tests/integration_test.rs
```diff
@@ -0,0 +1,174 @@
...
+    eth_pkt.set_src_mac(packets::MacAddr::new([0xac, 0x67, 0x3d, 0xfa, 0xbc, 0xdd]));
+    eth_pkt.set_dest_mac(packets::MacAddr::new([0xac, 0x67, 0x3d, 0xfa, 0xbc, 0xdf]));
+
+    eth_pkt.data
+}
+
+#[test]
+#[should_panic]
+#[ignore]
+fn output_panics_when_built_without_ingressor() {
```
CollinValley, line 56: remove ignore

route-rs-afpacket/tests/integration_test.rs
```diff
@@ -0,0 +1,174 @@
...
+
+    AfPacketOutputLink::new()
+        .ingressor(packet_generator)
+        .build_link();
+}
+
+#[test]
+#[should_panic]
+#[ignore]
+fn output_panics_when_built_with_multiple_ingressors() {
```
CollinValley, line 75: remove ignore

route-rs-runtime/src/utils/test/harness.rs
```diff
@@ -2,7 +2,10 @@ use crate::link::{Link, TokioRunnable};
 use crate::utils::test::packet_collectors::ExhaustiveCollector;
```
CollinValley, line 2: This whole file needs to be dropped. Was only meant for testing purposes.

route-rs-afpacket/tests/integration_test.rs
```diff
@@ -0,0 +1,174 @@
...
+        let mut external = afpacket::AsyncBoundSocket::from_interface(&iface_name).unwrap();
+
+        // set up link
+        let link = AfPacketInputLink::new().channel(rx_rtr).build_link();
+        let _external_task = tokio::spawn(async move {
+            // send packet from afpacket socket -> route-rs-afpacket
+            external.send(&ext_pkt_data).await.unwrap();
+        });
+
+        run_link(link).await
```
CollinValley: change to `test_link(link, timeout).await` I'll merge the new test_link() for you by EoW so you can merge this.

route-rs-afpacket/tests/integration_test.rs
```diff
@@ -0,0 +1,174 @@
...
+        let mut external = afpacket::AsyncBoundSocket::from_interface(&iface_name).unwrap();
+        external.set_promiscuous(true).unwrap();
+
+        // set up link
+        let link = AfPacketOutputLink::new()
+            .ingressor(immediate_stream(vec![pkt_data]))
+            .channel(tx_rtr)
+            .build_link();
+
+        run_link(link).await;
```
CollinValley: similar change to test_link as above.

CollinValley (commented):
No big changes needed, but doc comments are. Commented what to do with the run_link() function in your integration tests. Overall, the code looks real solid, good work.

CollinValley:
This is the commit that changes run_link to test link. https://github.com/route-rs/route-rs/commit/ebdbbaaf9f175a7de7e5b8aeecffaf1ee5d1734d

What I recommend you do, after fixing the above issues, is submit a PR against dev/pi-home-router. That way you have the latest changes while we wait to be ready to merge the runtime changes to master.  Then in a couple weeks we will be back to working off master.

moosingin3space:
Closing this so we can work on it as part of the PHR.

PR #203, opened 2020-02-20
Init Pi Home Router example, add first links

CollinValley:
the dev/pi-home-router is a development branch.

This first PR starts off with a series of commits that modify the Runtime and Packets libraries. Some of these are breaking changes on master, and as such graphgen tests are disabled. 

This also sets out a somewhat opinionated PHR layout, for people to begin building their links inside. I hope that it can accommodate as many needs as possible, but I expect rework to be required as more links get created. 

Also creates our first interface links, using Afpacket and the like. They are commented out until AfPacket is all merged.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,5 +14,6 @@ members = [
     "examples/trivial-identity",
     "examples/dns-interceptor",
     "examples/minimal-static-router",
+    "examples/pi-home-router",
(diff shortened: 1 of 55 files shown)


CollinValley:
Going to merge this to dev branch.  I still have plenty of work to do, but the main interface is there. If you are working on a part and need to know where to put your code, ping me.

PR #204, opened 2020-02-23
runtime: Create log processors

ddowl:
This overdue PR creates 2 processors `Log` and `FileLog` that don't transform packets, but have a side-effect of logging to an `impl Write` or to a `File`, respectively. 

`Log` uses a [`BufWriter`](https://doc.rust-lang.org/std/io/struct.BufWriter.html) to batch calls to the given `impl Write`.

`FileLog` is just a wrapper over the `Log` processor, but exposes a simpler interface that requires just a filename string, and instantiates `Log` with a `File` handle. 

The tests for these processors are similar, but I couldn't come up with a decent abstraction to eliminate the redundancy, so I've left some TODOs noting future cleanup work.

### DIFF
--- a/route-rs-runtime/src/processor/file_log.rs
+++ b/route-rs-runtime/src/processor/file_log.rs
@@ -1,0 +1,78 @@
+use crate::processor::{Log, Processor};
+use std::fmt::Debug;
+use std::fs::File;
+use std::marker::PhantomData;
+
+/// Wraps Log processor with a simpler File-specific interface
+/// You must provide a unique log filename.
+pub struct FileLog<A> {
+    phantom: PhantomData<A>,
+    log_processor: Log<A, File>,
+}
+
+impl<A> FileLog<A> {
+    pub fn new(name: &str) -> std::io::Result<FileLog<A>> {
+        Ok(FileLog {
+            phantom: PhantomData,
+            log_processor: Log::new(File::create(name)?),
+        })
+    }
+}
+
+impl<A: Send + Clone + Debug> Processor for FileLog<A> {
+    type Input = A;
+    type Output = A;
+
+    fn process(&mut self, packet: Self::Input) -> Option<Self::Output> {
+        self.log_processor.process(packet)
+    }
+}
+
+// TODO: Find common testing abstraction for logging processors
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::fs::{create_dir_all, remove_file};
+    use std::io::Read;
+    use std::path::Path;
+    use uuid::Uuid;
+
+    fn test_file_log_processor(packets: Vec<i32>, expected_log: &str) {
+        let log_dir = Path::new("test_logs");
+        let log_file = format!("{}.log", Uuid::new_v4());
+        let log_path = log_dir.join(log_file);
+        create_dir_all(log_dir).unwrap();
+
+        let mut proc = FileLog::new(log_path.to_str().unwrap()).unwrap();
+
+        let res_packets: Vec<i32> = packets
+            .clone()
+            .into_iter()
+            .map(|packet| proc.process(packet).unwrap())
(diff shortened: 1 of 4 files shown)


route-rs-runtime/src/processor/log.rs
```diff
@@ -0,0 +1,95 @@
...
+    }
+}
+
+impl<A: Send + Clone + Debug, W: Write> Processor for Log<A, W> {
+    type Input = A;
+    type Output = A;
+
+    fn process(&mut self, packet: Self::Input) -> Option<Self::Output> {
+        self.log_writer
+            .write_all(format!("{:?}\n", packet).as_ref())
```
CollinValley, line 38: Good idea to just write the debug message.  I'm not sure how well formatted the packets library packets are, but that could always be improved

route-rs-runtime/src/processor/mod.rs
```diff
@@ -19,6 +19,12 @@ pub use self::drop::*;
 mod dec_ip_hop;
 pub use self::dec_ip_hop::*;
 
+mod log;
+pub use self::log::*;
```
CollinValley, line 23: Nit: In the future, let's not do this with a wildcard, we should be only exporting the parts that are the interface.  In this case, the only struct there is an interface, so this is the same anyways. Up to you if you want to rectify.

route-rs-runtime/src/processor/file_log.rs
```diff
@@ -0,0 +1,78 @@
...
+
+/// Wraps Log processor with a simpler File-specific interface
+/// You must provide a unique log filename.
+pub struct FileLog<A> {
+    phantom: PhantomData<A>,
+    log_processor: Log<A, File>,
+}
+
+impl<A> FileLog<A> {
+    pub fn new(name: &str) -> std::io::Result<FileLog<A>> {
```
CollinValley, line 14: Good job using appropriate result here.  I guess Rust won't let you not get the types correct, but still, good stuff.

route-rs-runtime/src/processor/file_log.rs
```diff
@@ -0,0 +1,78 @@
+use crate::processor::{Log, Processor};
+use std::fmt::Debug;
+use std::fs::File;
+use std::marker::PhantomData;
+
+/// Wraps Log processor with a simpler File-specific interface
```
CollinValley, line 6: Nit: Add more documentation here. Maybe something about how the file will be created, not appended to, can't exist already, etc.  Also talk about how we use buffered writes to keep it fast? Just some suggestions.

route-rs-runtime/src/processor/file_log.rs
```diff
@@ -0,0 +1,78 @@
...
+mod tests {
+    use super::*;
+    use std::fs::{create_dir_all, remove_file};
+    use std::io::Read;
+    use std::path::Path;
+    use uuid::Uuid;
+
+    fn test_file_log_processor(packets: Vec<i32>, expected_log: &str) {
+        let log_dir = Path::new("test_logs");
+        let log_file = format!("{}.log", Uuid::new_v4());
```
CollinValley, line 42: This is a good solution for creating throwaway files.

CollinValley (approved):
Couple nits, but LGTMMD!

route-rs-runtime/src/processor/mod.rs
ddowl: We'll probably go back and fix them all in one sweep, but I'll keep that in mind for next time.

route-rs-runtime/src/processor/file_log.rs
ddowl: I'll make a short follow up change to add some better docs. It'd be nice to explain why we're using `BufWriter` since it seems like a great abstraction to use here.

PR #206, opened 2020-02-29
PHR IPv4

scgruber:
This is a working branch for the IPv4 implementation. Will be splitting the commits into more logical pieces and adding tests before this is okay to merge in.

At this point I think the structure of the link is complete though.

### DIFF
--- a/examples/pi-home-router/src/interface/classifier/inbound_interface.rs
+++ b/examples/pi-home-router/src/interface/classifier/inbound_interface.rs
@@ -1,0 +1,25 @@
+use crate::types::{Interface, InterfaceAnnotated};
+use route_rs_packets::Packet;
+use route_rs_runtime::classifier::Classifier;
+use std::marker::PhantomData;
+
+pub(crate) struct ByInboundInterface<P: Send + Clone + Packet> {
+    phantom: PhantomData<P>,
+}
+
+impl<P: Send + Clone + Packet> ByInboundInterface<P> {
+    pub(crate) fn new() -> Self {
+        ByInboundInterface {
+            phantom: PhantomData,
+        }
+    }
+}
+
+impl<P: Send + Clone + Packet> Classifier for ByInboundInterface<P> {
+    type Packet = InterfaceAnnotated<P>;
+    type Class = Interface;
+
+    fn classify(&self, packet: &Self::Packet) -> Self::Class {
+        packet.inbound_interface
+    }
+}
--- a/examples/pi-home-router/src/interface/link/interface_collect.rs
+++ b/examples/pi-home-router/src/interface/link/interface_collect.rs
@@ -5,7 +5,7 @@ use route_rs_runtime::link::{
     primitive::{JoinLink, ProcessLink},
     Link, LinkBuilder, PacketStream, ProcessLinkBuilder,
 };
-use route_rs_runtime::unpack;
+use route_rs_runtime::unpack_link;
 
 /// InterfaceCollect: Link that our 3 input interfaces, Host, Lan and Wan, expected in that order.
 pub(crate) struct InterfaceCollect {
@@ -48,35 +48,32 @@ impl LinkBuilder<EthernetFrame, InterfaceAnnotated<EthernetFrame>> for Interface
     fn build_link(self) -> Link<InterfaceAnnotated<EthernetFrame>> {
         let mut streams = self.in_streams.unwrap();
 
-        unpack!(
-            let (_, [host]) = ProcessLink::new()
-                .processor(InterfaceAnnotationEncap::new(
-                    Interface::Host,
-                    Interface::Unmarked
-                ))
-                .ingressor(streams.remove(0))
-                .build_link();
-        );
+        let annotate_host = ProcessLink::new()
+            .processor(InterfaceAnnotationEncap::new(
+                Interface::Host,
+                Interface::Unmarked,
+            ))
+            .ingressor(streams.remove(0))
+            .build_link();
+        unpack_link!(annotate_host => _, [host]);
 
-        unpack!(
-            let (_, [lan]) = ProcessLink::new()
-                .processor(InterfaceAnnotationEncap::new(
-                    Interface::Lan,
-                    Interface::Unmarked
-                ))
-                .ingressor(streams.remove(1))
-                .build_link();
-        );
+        let annotate_lan = ProcessLink::new()
+            .processor(InterfaceAnnotationEncap::new(
+                Interface::Lan,
+                Interface::Unmarked,
+            ))
+            .ingressor(streams.remove(1))
+            .build_link();
+        unpack_link!(annotate_lan => _, [lan]);
 
-        unpack!(
-            let (_, [wan]) = ProcessLink::new()
-                .processor(InterfaceAnnotationEncap::new(
-                    Interface::Wan,
-                    Interface::Unmarked
-                ))
-                .ingressor(streams.remove(0))
-                .build_link();
-        );
+        let annotate_wan = ProcessLink::new()
+            .processor(InterfaceAnnotationEncap::new(
+                Interface::Wan,
+                Interface::Unmarked,
+            ))
+            .ingressor(streams.remove(0))
+            .build_link();
+        unpack_link!(annotate_wan => _, [wan]);
 
         // Join link has the only tokio runnables here, can just return it
         JoinLink::new()
--- a/examples/pi-home-router/src/interface/link/interface_dispatch.rs
+++ b/examples/pi-home-router/src/interface/link/interface_dispatch.rs
@@ -71,21 +71,21 @@ impl LinkBuilder<InterfaceAnnotated<EthernetFrame>, EthernetFrame> for Interface
         unpack!(
             let (_, [host_e]) = ProcessLink::new()
                 .ingressor(ce_0)
-                .processor(InterfaceAnnotationDecap)
+                .processor(InterfaceAnnotationDecap::new())
                 .build_link();
         );
 
         unpack!(
             let (_, [lan_e]) = ProcessLink::new()
                 .ingressor(ce_1)
-                .processor(InterfaceAnnotationDecap)
+                .processor(InterfaceAnnotationDecap::new())
                 .build_link();
         );
 
         unpack!(
             let (_, [wan_e]) = ProcessLink::new()
                 .ingressor(ce_2)
-                .processor(InterfaceAnnotationDecap)
+                .processor(InterfaceAnnotationDecap::new())
                 .build_link();
         );
 
--- a/examples/pi-home-router/src/interface/link/interface_dispatch.rs
+++ b/examples/pi-home-router/src/interface/link/interface_dispatch.rs
@@ -6,7 +6,7 @@ use route_rs_runtime::link::{
     primitive::{ClassifyLink, ProcessLink},
     Link, LinkBuilder, PacketStream, ProcessLinkBuilder,
 };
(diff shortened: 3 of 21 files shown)


examples/pi-home-router/src/interface/link/interface_collect.rs
```diff
@@ -89,7 +89,7 @@ mod tests {
         let packets = vec![EthernetFrame::empty(); 3];
         let host = immediate_stream(packets.clone());
         let lan = immediate_stream(packets.clone());
-        let wan = immediate_stream(packets.clone());
+        let wan = immediate_stream(packets);
```
CollinValley: Good catch

examples/pi-home-router/src/interface/link/interface_dispatch.rs
```diff
@@ -71,17 +71,17 @@ impl LinkBuilder<InterfaceAnnotated<EthernetFrame>, EthernetFrame> for Interface
 
         let (_, host_e) = ProcessLink::new()
             .ingressor(ce.remove(0))
-            .processor(InterfaceAnnotationDecap)
+            .processor(InterfaceAnnotationDecap::new())
```
CollinValley, line 72: Unfortunate that we need the PhantomData with generic structs. But it makes sense.

examples/pi-home-router/src/interface/processor/annotation_set_out.rs
```diff
@@ -0,0 +1,35 @@
+use crate::types::{Interface, InterfaceAnnotated};
+use route_rs_packets::Packet;
+use route_rs_runtime::processor::Processor;
+use std::marker::PhantomData;
+
+/// InterfaceAnnotationEncap: Processor to apply interface annotation to a packet
```
CollinValley: Need to update the doc comment here.

examples/pi-home-router/src/ipv4/mod.rs
```diff
@@ -0,0 +1,388 @@
...
+pub(crate) struct HandleIPv4 {
+    in_stream: Option<PacketStream<InterfaceAnnotated<Ipv4Packet>>>,
+    wan_ip: Option<Ipv4Addr>,
+    lan_ip: Option<Ipv4Addr>,
+    lan_subnet: Option<Ipv4Cidr>,
+}
+
+#[allow(dead_code)]
+impl HandleIPv4 {
+    pub(crate) fn new() -> Self {
```
CollinValley, line 31: Thinking about how configuration will work in our router, from the top level. Perhaps we will need to create some sort of configuration structure.  Think how C libs commonly take a bunch of arguments, you make a structure that gets passed in, that struct is used to set opts.  Perhaps we could also just accept a JSON object as an argument or something?  Nothing to worry about for now, something to consider if the number of config options gets high.

examples/pi-home-router/src/interface/processor/annotation_decap.rs
```diff
@@ -1,14 +1,24 @@
 use crate::types::InterfaceAnnotated;
-use route_rs_packets::EthernetFrame;
+use route_rs_packets::Packet;
```
CollinValley, line 2: Thanks for expanding this to be generic for me.

examples/pi-home-router/src/interface/processor/annotation_encap.rs
```diff
@@ -1,28 +1,31 @@
 use crate::types::{Interface, InterfaceAnnotated};
-use route_rs_packets::EthernetFrame;
+use route_rs_packets::Packet;
```
CollinValley, line 2: Same as above: 👍

examples/pi-home-router/src/interface/processor/annotation_set_out.rs
CollinValley: Consider also using a different shortening for output.  I was a bit confused by the name when I saw it.

examples/pi-home-router/src/ipv4/mod.rs
```diff
@@ -0,0 +1,388 @@
...
+                in_stream: Some(in_stream),
+                wan_ip: self.wan_ip,
+                lan_ip: self.lan_ip,
+                lan_subnet: self.lan_subnet,
+            },
+            Some(_) => panic!("HandleIPv4 takes only one ingressor"),
+        }
+    }
+
+    fn build_link(self) -> Link<InterfaceAnnotated<Ipv4Packet>> {
```
CollinValley, line 106: You should consider breaking this into more than just one mega-link.  Ideally each build_link, like a function, doesn't get this long.  I know this is a draft so it is fine right now.

examples/pi-home-router/src/ipv4/mod.rs
```diff
@@ -0,0 +1,388 @@
...
+        let (mut nat_encap_runnables, mut nat_encap_egressors) = ProcessLink::new()
+            .ingressor(from_join_host_lan_for_nat)
+            .processor(Identity::new())
+            .build_link();
+        all_runnables.append(&mut nat_encap_runnables);
+        let from_nat_encap: PacketStream<Ipv4Packet> = nat_encap_egressors.remove(0);
+
+        // Since all NAT encapsulated traffic is from the LAN bound for the WAN, we can just
+        // regenerate the interface annotations here. Traffic from the Host is marked as LAN here,
+        // but that's not really inaccurate and we won't use this information later anyway.
```
CollinValley: Perhaps just leave the Inbound_inteface as unmarked then? To convey to the downstream that it is unknown information as this point.  You could also probably rebuild it (host vs LAN) just by checking the src ip of the packet.

examples/pi-home-router/src/ipv4/mod.rs
```diff
@@ -0,0 +1,388 @@
...
+            ProcessLink::new()
+                .ingressor(from_wan_to_me_annotated)
+                .processor(InterfaceAnnotationDecap::new())
+                .build_link();
+        all_runnables.append(&mut wan_to_me_annote_decap_runnables);
+        let from_wan_to_me: PacketStream<Ipv4Packet> = wan_to_me_annot_decap_egressors.remove(0);
+
+        // Send traffic from the WAN to the NAT decapsulator.
+        //
+        // TODO: Implement NAT decapsulator
```
CollinValley, line 297: Decapsulator is a good name, sounds like an evil villain weapon.

examples/pi-home-router/src/ipv4/mod.rs
CollinValley: Ideally we don't ever add incorrect information to a packet.

CollinValley (commented):
Overall, this is really exciting, awesome work, thanks for your help!

examples/pi-home-router/src/interface/link/interface_collect.rs
scgruber: I can only praise our lord and savior clippy for this one.

examples/pi-home-router/src/ipv4/mod.rs
scgruber: Yeah I think figuring out the right balance between explicitness and ease of passing configuration is going to be pretty important, and I'm not sure exactly what to do about that yet.

examples/pi-home-router/src/ipv4/mod.rs
scgruber: I don't think this is actually that complex of a logic, there are maybe 20 links in here. The problem I think is that setting up each element is this huge block even though conceptually there's not that much going on.

Personally I don't like trying to split things into so many pieces just for the goal of splitting them up, because it can become much harder to follow the flow of what's happening.

This is also an area where I expect the graph view to be much easier to interact with. I sketched a graph on paper when I was building this and it was pretty straightforward to follow, even though linearizing it in code is less clear.

examples/pi-home-router/src/interface/classifier/inbound_interface.rs
```diff
@@ -0,0 +1,30 @@
...
+        }
+    }
+}
+
+impl<P: Send + Clone + Packet> Classifier for ByInboundInterface<P> {
+    type Packet = InterfaceAnnotated<P>;
+    type Class = Interface;
+
+    fn classify(&self, packet: &Self::Packet) -> Self::Class {
+        match packet.inbound_interface {
```
ddowl: just return `packet.inbound_interface`.





bruh

examples/pi-home-router/src/ipv4/mod.rs
```diff
@@ -0,0 +1,473 @@
...
+        all_runnables.append(&mut classify_by_inbound_interface_runnables);
+        let from_host: PacketStream<InterfaceAnnotated<Ipv4Packet>> =
+            classify_by_inbound_interface_egressors.remove(0);
+        let from_lan: PacketStream<InterfaceAnnotated<Ipv4Packet>> =
+            classify_by_inbound_interface_egressors.remove(0);
+        let from_wan: PacketStream<InterfaceAnnotated<Ipv4Packet>> =
+            classify_by_inbound_interface_egressors.remove(0);
+
+        // Packets from the Host either are bound for the LAN subnet (or LAN broadcast), else they
+        // go out to the WAN. Broadcast packets shouldn't be sent up to the LAN because it is a
```
ddowl: oops this should be "sent up to the WAN"

examples/pi-home-router/src/ipv4/classify_dest_subnet.rs
```diff
@@ -0,0 +1,45 @@
...
+use route_rs_packets::Ipv4Packet;
+use route_rs_runtime::classifier::Classifier;
+use std::collections::HashMap;
+
+/// Classifies an Ipv4Packet by its destination subnet.
+///
+/// This classifier uses a HashMap of CIDRs to subnet names. We require but do not enforce that the
+/// CIDRs are disjoint. Undefined behavior will occur if they are overlapping.
+///
+/// TODO: Actually enforce that CIDRs are disjoint at runtime.
```
ddowl: Actually let's allow overlaps but prefer the longest CIDR to match. We can sort the keys by length once in the constructor for performance.

route-rs-runtime/src/lib.rs
```diff
@@ -65,3 +65,34 @@ macro_rules! unpack {
         unpack!(HELPER $vector, $($y), +)
     };
 }
+
+#[macro_export]
+macro_rules! unpack_link {
```
CollinValley, line 37: This looks good, I like the benefits of this newer layout.  However can you make this one supplant the old one?  Choose either name.

examples/pi-home-router/src/interface/processor/annotation_set_inbound.rs
```diff
@@ -0,0 +1,34 @@
+use crate::types::{Interface, InterfaceAnnotated};
+use route_rs_packets::Packet;
+use route_rs_runtime::processor::Processor;
+use std::marker::PhantomData;
+
+/// InterfaceAnnotationSetOutbound: Update the outbound interface for a packet
```
CollinValley: Doc comment needs updating.

CollinValley (changes requested):
Just a few final things, then I think this is ready for merge.  I really like the trouble you've went to to provide good test coverage as well, makes the code a lot more professional.

route-rs-runtime/src/lib.rs
scgruber: Yeah I just didn't want to go through all the trouble of converting everywhere before someone else approved it. Will update things now.

scgruber:
Should be good to go now

CollinValley (approved):

PR #207, opened 2020-03-01
PHR: Rework Intake and Exhaust Links

CollinValley:
At our last meeting, other members brought up that we should not code
the IO layers into the business logic of our core router, particularly
since we want to be able to test the router using PCAPs, and swapping
real OS interfaces for PCAP streams shouldn't be a pain. So I am
reworking the intake and exhaust to not do this, and to handle
converting the raw Vec<u8> packet streams to route-rs Packets.

Note: This also makes the currently created interface sections no longer dependent on AfPacket.  I'll create a new link to handle creation and deletion of the AfPacket links once they are merged. Ideally we will be able to handle hooking up PCAPs the same way, so testing is simplified.


CollinValley:
Merging this, there isn't anything too fancy here, and it is just the dev branch.

PR #208, opened 2020-03-02
travis.yml: Test dev/ branches

scgruber:


CollinValley:
Does this make sense? On the one hand, it is good to always pass tests, on the other, this is a development branch. It exists to sidestep the tooling on master to allow for more flexible iteration. I'll accept this for now but if it become a headache I'll remove it.

scgruber:
The reason I wanted to add this is I want to be able to run tests locally when I am working on my parts, and it's really frustrating if all of the other tests are broken when I'm trying to do that. We can always `#[ignore]` tests that don't pass because we're in the middle of development.

PR #209, opened 2020-03-04
PHR ARP

ddowl:
This draft contains skeleton implementations of 2 important components specified in ARP's [RFC](https://tools.ietf.org/html/rfc826), handling requests (`ArpHandler`), and generating requests (`ArpGenerator`). There are several issues at the moment:

- I've defined the interface of an `ArpFrame` wrapper around `EthernetFrame` that provides some helper functions to read/write values in an ARP frame. The Links are using the interface, but the functions are unimplemented.
- I've defined an interface to various translation tables called `ArpTable`, but it isn't global or locked.
- There are open questions about where `ArpFrame` and/or `arp_utils` belong (PHR or route-rs-packets).
- How to look up local IP config data. (What is the router's IP address?)
- Poor abstraction over IPv4/IPv6 handling. Since ARP looks for `(protocol type, protocol addr)` tuples, there's a lot of duplicated code performing similar actions depending on network protocol (particularly in `ArpGenerator`). We should find a better way to handle these cases. 
- No tests.

### DIFF
--- a/examples/pi-home-router/src/arp/arp_frame.rs
+++ b/examples/pi-home-router/src/arp/arp_frame.rs
@@ -1,0 +1,88 @@
+use route_rs_packets::{EthernetFrame, MacAddr};
+use std::net::IpAddr;
+
+pub(crate) enum ArpOp {
+    Request = 1,
+    Reply = 2,
+}
+
+pub(crate) enum ArpHardwareType {
+    Ethernet = 1,
+}
+
+pub(crate) const ARP_ETHER_TYPE: u16 = 0x0806;
+
+// NOTE: Could be implemented in various ways, such as a specialized version of EthernetFrame that's
+// known to be an ARP frame. It could be implemented in a similar way that packets are promoted/demoted
+// with `TryFrom`.
+#[derive(Clone)]
+pub(crate) struct ArpFrame {
+    frame: EthernetFrame,
+}
+
+impl ArpFrame {
+    pub fn new(frame: EthernetFrame) -> Self {
+        assert_eq!(frame.ether_type(), ARP_ETHER_TYPE);
+        ArpFrame { frame }
+    }
+
+    pub fn hardware_type(&self) -> u16 {
+        unimplemented!()
+    }
+
+    pub fn protocol_type(&self) -> u16 {
+        unimplemented!()
+    }
+
+    pub fn hardware_addr_len(&self) -> u8 {
+        unimplemented!()
+    }
+
+    pub fn protocol_addr_len(&self) -> u8 {
+        unimplemented!()
+    }
+
+    pub fn opcode(&self) -> u8 {
+        unimplemented!()
+    }
+
+    pub fn sender_hardware_addr(&self) -> &[u8] {
+        unimplemented!()
+    }
+
+    pub fn sender_protocol_addr(&self) -> &[u8] {
+        unimplemented!()
+    }
+
+    pub fn target_hardware_addr(&self) -> &[u8] {
+        unimplemented!()
+    }
+
+    pub fn target_protocol_addr(&self) -> &[u8] {
+        unimplemented!()
+    }
+
+    pub fn set_opcode(&mut self, code: u8) {
+        unimplemented!()
+    }
+
+    pub fn set_sender_hardware_addr(&mut self, addr: MacAddr) {
+        unimplemented!()
+    }
+
+    pub fn set_sender_protocol_addr(&mut self, ip_addr: IpAddr) {
+        unimplemented!()
+    }
+
+    pub fn set_target_hardware_addr(&mut self, addr: MacAddr) {
+        unimplemented!()
+    }
+
+    pub fn set_target_protocol_addr(&mut self, ip_addr: IpAddr) {
+        unimplemented!()
+    }
+
+    pub fn frame(self) -> EthernetFrame {
+        self.frame
+    }
+}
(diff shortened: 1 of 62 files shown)


examples/pi-home-router/src/arp/arp_frame.rs
```diff
@@ -0,0 +1,106 @@
...
+    Reply = 2,
+}
+
+pub(crate) enum ArpHardwareType {
+    Ethernet = 1,
+}
+
+pub(crate) const ARP_ETHER_TYPE: u16 = 0x0806;
+
+// NOTE: Could be implemented in various ways, such as a specialized version of EthernetFrame that's
```
scgruber, line 15: I think this is the reasonable way to do this. We should put it in route-rs-packets rather than asking every implementor to write their own ARP frame type.

examples/pi-home-router/src/arp/arp_generator.rs
```diff
@@ -0,0 +1,114 @@
...
+        }
+    }
+}
+
+impl Processor for ArpGenerator {
+    type Input = InterfaceAnnotated<EthernetFrame>;
+    type Output = InterfaceAnnotated<EthernetFrame>;
+
+    ///
+    /// As a packet is sent down through the network layers, routing determines the protocol address
```
scgruber, line 27: This seems like it's coming from an RFC or something, can you cite the source?

examples/pi-home-router/src/arp/arp_generator.rs
```diff
@@ -0,0 +1,114 @@
...
+    /// It does not set Target Hardware Address to anything in particular, because it is this value
+    /// that it is trying to determine. It could set the Target Hardware Address to the broadcast
+    /// address (FF:FF:FF:FF:FF:FF) for the hardware (all ones in the case of the 10Mbit Ethernet)
+    /// if that makes it convenient for some aspect of the implementation.
+    ///
+    /// It then causes this packet to be broadcast to all stations on the Ethernet cable originally
+    /// determined by the routing mechanism.
+    ///
+    fn process(&mut self, packet: Self::Input) -> Option<Self::Output> {
+        // It would be nice if there was a function to determine which network protocol type
```
scgruber, line 57: You should be able to match on `frame.ether_type()`. Probably we could add some human-readable names in router-rs-packets like we do for `IpProtocol`.

examples/pi-home-router/src/arp/arp_generator.rs
```diff
@@ -0,0 +1,114 @@
...
+use crate::arp::{ArpFrame, ArpTable, IPV4_PROTOCOL_TYPE};
+use crate::types::InterfaceAnnotated;
+use route_rs_packets::{EthernetFrame, Ipv4Packet, Ipv6Packet, MacAddr};
+use route_rs_runtime::processor::Processor;
+use std::convert::TryFrom;
+use std::net::IpAddr;
+
+pub(crate) struct ArpGenerator {
+    // Mapping between (protocol type, protocol address) -> 48-bit Ethernet address
+    arp_table: ArpTable,
```
scgruber, line 11: This probably needs to be a reference

examples/pi-home-router/src/arp/arp_handler.rs
```diff
@@ -0,0 +1,133 @@
...
+use route_rs_runtime::processor::Processor;
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
+
+// TODO: We need to respond to ARP requests that are targeted at the router.
+// Where is this state stored? How to we access it?
+const ROUTER_IPV4_ADDR: Ipv4Addr = Ipv4Addr::LOCALHOST;
+const ROUTER_IPV6_ADDR: Ipv6Addr = Ipv6Addr::LOCALHOST;
+
+pub(crate) struct ArpHandler {
+    arp_table: ArpTable,
```
scgruber, line 16: This probably needs to be a mutable reference

examples/pi-home-router/src/arp/arp_table.rs
```diff
@@ -0,0 +1,61 @@
...
+    pub fn new() -> Self {
+        ArpTable {
+            ipv4_mac_translations: HashMap::new(),
+            ipv6_mac_translations: HashMap::new(),
+        }
+    }
+
+    pub fn contains_key(&self, protocol_type: u16, protocol_address: &[u8]) -> bool {
+        match protocol_type {
+            0x0800 => self
```
scgruber, line 23: There's probably a cleaner way to switch on the table type without duplicating the method calls every time.

examples/pi-home-router/src/arp/arp_generator.rs
ddowl: I stuck it in the PR description, but I can put it with the code as well. `ArpHandler` and `ArpGenerator` both have paraphrased descriptions from the RFC.

examples/pi-home-router/src/arp/arp_generator.rs
ddowl: That's a good point. I defined constants for those in `arp_utils`, but I think they should live in the packet library.

examples/pi-home-router/src/arp/arp_table.rs
ddowl: Yeeeeeeeep

ddowl:
Closing, as I've submitted a more complete, non-draft PR [here](https://github.com/route-rs/route-rs/pull/224).

PR #210, opened 2020-03-05
FromPcap/ToPcap: add support to read/write pcap

ErrHu:
Add the ability to read/write packets from pcap.
FromPcap offers two mode: Dump and Realtime
Dump - Pass next packets whenever polled
Realtime - Pass packets once their timestamp has passed

This was tested with a pcap of a ping to 8.8.8.8

For issue: #173


route-rs-runtime/src/link/primitive/pcap.rs
```diff
@@ -0,0 +1,377 @@
...
+    fn poll(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
+        let collector = Pin::into_inner(self);
+        loop {
+            match ready!(Pin::new(&mut collector.stream).poll_next(cx)) {
+                Some(packet) => {
+                    let ts = collector.start.elapsed();
+                    let tv_sec = ts.as_secs();
+                    let tv_usec = (ts - Duration::from_secs(tv_sec)).as_micros();
+                    let packet_hdr = PacketHeader {
+                        ts: libc::timeval {
```
CollinValley, line 261: Hey Eric, I'm getting a build error here saying that this should be an i32, not an i64
```
error[E0308]: mismatched types
   --> route-rs-runtime/src/link/primitive/pcap.rs:264:38
    |
264 | ...                   tv_usec: tv_usec as i64,
    |                                ^^^^^^^^^^^^^^ expected `i32`, found `i64`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0308`.
error: could not compile `route-rs-runtime`.
```

CollinValley: May be possible this is an OS-specific issue? Changing that to be an as i32 resolves the error on my end, all the tests pass then.

route-rs-runtime/src/link/primitive/pcap.rs
```diff
@@ -0,0 +1,377 @@
+//use crate::link::Link;
```
CollinValley: Nit: Remove extraneous comment

route-rs-runtime/src/link/primitive/pcap.rs
CollinValley: LOL it is OS-specific.  When I `cargo build --target x86_64-unknown-linux-gnu` it works, doesn't on MacOS.  Could you add a cfg flag for MacOS to be i32 instead of i64? The runtime still supports MacOS somewhat, it should at least build on it.

CollinValley: BTW to add targets, you just `rustup target add <triple>`.  List of supported triples are here https://forge.rust-lang.org/release/platform-support.html

CollinValley: Graphgen doesn't cross-compile for me right away, I think it needs a C linker for some reason, it threw a wall of text at me in classic linker failure fashion.  But I can cross-compile the runtime for windows, macOS, and Linux. 

Here is a patch that sets the conditional compilation flags appropriately

CollinValley: [0001-Runtime-PCAP-Fix-compilation-for-MacOS.patch.txt](https://github.com/route-rs/route-rs/files/4291531/0001-Runtime-PCAP-Fix-compilation-for-MacOS.patch.txt)

CollinValley (changes requested):
This is pretty much ready for merge.  Travis is being weird, so I pulled the code and tested locally. Clippy errors are resolved.  If you apply my patch and push that, then this is ready for merge.

route-rs-runtime/src/link/primitive/pcap.rs
ErrHu: removed

route-rs-runtime/src/link/primitive/pcap.rs
ErrHu: Applied the patch and made a comment in the git log

CollinValley (commented):
LGTMMD

CollinValley (approved):

PR #212, opened 2020-03-08
packets: Add ArpFrame wrapper around EthernetFrame

ddowl:
Wanted to get this into master since `ArpFrame` is independent of PHR. This change introduces a wrapper of `EthernetFrame` with getters and setters for the payload schema of an ARP frame, found in [its RFC](https://tools.ietf.org/html/rfc826). The implementation is fairly straightforward, but I've highlighted some questions/improvements in TODO comments:

- ARP frame generation: In order to construct an `EthernetFrame` of the appropriate length, we need the size of the hardware address and protocol address, which will almost always refer to Ethernet (6 bytes) and IPv4 (4 bytes). Personally, I'd like to construct an interface that doesn't leave any in-between states, or partially constructed ARP frames (see `generate_empty_arp_frame` test). Maybe we can just accept a struct of options to initialize an `ArpFrame`, or maybe use a builder pattern like the rest of the runtime?
- I find myself referencing an `EthernetFrame`'s payload via its `data` field directly, rather than its `payload()` function due to ownership problems. I'd either need some advice working with `Cow`s or change the signature of `EthernetFrame::payload()`.
- The only ether types we probably care about are IPv4, IPv6, and ARP. Should those `const` definitions live in `ethernet.rs` or `types.rs`?

### DIFF
--- a/route-rs-packets/src/arp.rs
+++ b/route-rs-packets/src/arp.rs
@@ -212,7 +212,7 @@ impl TryFrom<EthernetFrame> for ArpFrame {
     ///
     fn try_from(frame: EthernetFrame) -> Result<Self, Self::Error> {
         if frame.ether_type() != ARP_ETHER_TYPE {
-            return Err("Frame does not have ARP ether type.");
+            return Err("Frame does not have ARP ether type");
         };
 
         let arp_frame = ArpFrame { frame };
@@ -267,7 +267,7 @@ mod tests {
     }
 
     #[test]
-    fn arp_frame_from_ethernet() -> Result<(), String> {
+    fn arp_frame_from_ethernet() -> Result<(), &'static str> {
         let arp_payload: Vec<u8> = vec![
             0x00, 0x01, 0x00, 0x01, 0x06, 0x04, 0x00, 0x01, 1, 2, 3, 4, 5, 6, 10, 0, 0, 1, 10, 9,
             8, 7, 6, 5, 0xff, 0xff, 0xff, 0xff,
@@ -287,5 +287,21 @@ mod tests {
         assert_eq!(arp_frame.target_hardware_addr(), [10, 9, 8, 7, 6, 5]);
         assert_eq!(arp_frame.target_protocol_addr(), [0xff, 0xff, 0xff, 0xff]);
         Ok(())
+    }
+
+    #[test]
+    #[should_panic(expected = "Frame does not have ARP ether type")]
+    fn try_from_non_arp_ether_type() {
+        let mut ethernet_frame = EthernetFrame::empty();
+        ethernet_frame.set_ether_type(ARP_ETHER_TYPE + 1);
+        ArpFrame::try_from(ethernet_frame).unwrap();
+    }
+
+    #[test]
+    #[should_panic(expected = "Frame payload is too small")]
+    fn try_from_small_frame() {
+        let mut ethernet_frame = EthernetFrame::empty();
+        ethernet_frame.set_ether_type(ARP_ETHER_TYPE);
+        ArpFrame::try_from(ethernet_frame).unwrap();
     }
 }
--- a/route-rs-packets/src/arp.rs
+++ b/route-rs-packets/src/arp.rs
@@ -1,5 +1,5 @@
-use crate::{EthernetFrame, MacAddr};
-use std::convert::TryInto;
+use crate::{EthernetFrame, MacAddr, ARP_ETHER_TYPE};
+use std::convert::{TryFrom, TryInto};
 use std::net::IpAddr;
 
 pub enum ArpOp {
@@ -11,11 +11,6 @@ pub enum ArpHardwareType {
     Ethernet = 1,
 }
 
-// TODO: should move to types.rs or ethernet.rs
-pub const IPV4_ETHER_TYPE: u16 = 0x0800;
-pub const IPV6_ETHER_TYPE: u16 = 0x86DD;
-pub const ARP_ETHER_TYPE: u16 = 0x0806;
-
 const HARDWARE_TYPE_RANGE: (usize, usize) = (0, 2);
 const PROTOCOL_TYPE_RANGE: (usize, usize) = (2, 4);
 const HARDWARE_ADDR_LEN_RANGE: (usize, usize) = (4, 5);
@@ -26,17 +21,16 @@ const OPCODE_RANGE: (usize, usize) = (6, 8);
 /// EthernetFrame wrapper with getters/setters for the packet structure described in RFC 826
 /// https://tools.ietf.org/html/rfc826
 ///
-// NOTE: Could be implemented in various ways, such as a specialized version of EthernetFrame that's
-// known to be an ARP frame. It could be implemented in a similar way that packets are promoted/demoted
-// with `TryFrom`.
 #[derive(Clone)]
 pub struct ArpFrame {
     frame: EthernetFrame,
 }
 
 impl ArpFrame {
-    // TODO: would like default args for Ethernet (6 bytes) and IPv4 (4 bytes)
-    // TODO: Should we accept a struct of options to initialize the ArpFrame with? Or use a builder pattern?
+    ///
+    /// Constructs a new, empty packet with a payload big enough for all ARP fields,
+    /// given some hardware/protocol address lengths.
+    ///
     pub fn new(hardware_addr_len: u8, protocol_addr_len: u8) -> Self {
         let payload_len = 8 + (2 * hardware_addr_len as usize) + (2 * protocol_addr_len as usize);
         let payload: Vec<u8> = vec![0; payload_len];
@@ -48,11 +42,6 @@ impl ArpFrame {
         arp_frame.set_hardware_addr_len(hardware_addr_len);
         arp_frame.set_protocol_addr_len(protocol_addr_len);
         arp_frame
-    }
-
-    pub fn from(frame: EthernetFrame) -> Self {
-        assert_eq!(frame.ether_type(), ARP_ETHER_TYPE);
-        ArpFrame { frame }
     }
 
     pub fn hardware_type(&self) -> u16 {
@@ -211,6 +200,38 @@ impl ArpFrame {
     }
 }
 
+impl TryFrom<EthernetFrame> for ArpFrame {
+    type Error = &'static str;
+
+    ///
+    /// Decorates the given EthernetFrame with ArpFrame getters/setters.
+    /// Validates
+    /// - The frame has an ARP ether type
+    /// - The frame has a reasonable payload size given the hardware/protocol address lengths
+    ///
+    fn try_from(frame: EthernetFrame) -> Result<Self, Self::Error> {
+        if frame.ether_type() != ARP_ETHER_TYPE {
+            return Err("Frame does not have ARP ether type.");
+        };
+
+        let arp_frame = ArpFrame { frame };
+        let payload_len = arp_frame.frame.payload().len();
+
+        if payload_len < 8 {
+            return Err("Frame payload is too small");
+        }
+
+        let hlen = arp_frame.hardware_addr_len() as usize;
+        let plen = arp_frame.protocol_addr_len() as usize;
+
+        if payload_len != (8 + (2 * hlen) + (2 * plen)) {
+            return Err("Frame payload doesn't match address length fields");
+        }
+
+        Ok(arp_frame)
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
@@ -230,7 +251,7 @@ mod tests {
     }
 
     #[test]
-    fn arp_frame_from_ethernet() {
+    fn arp_frame_from_ethernet() -> Result<(), String> {
         let arp_payload: Vec<u8> = vec![
             0x00, 0x01, 0x00, 0x01, 0x06, 0x04, 0x00, 0x01, 1, 2, 3, 4, 5, 6, 10, 0, 0, 1, 10, 9,
             8, 7, 6, 5, 0xff, 0xff, 0xff, 0xff,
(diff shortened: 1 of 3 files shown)


route-rs-packets/src/arp.rs
```diff
@@ -0,0 +1,253 @@
...
+///
+/// EthernetFrame wrapper with getters/setters for the packet structure described in RFC 826
+/// https://tools.ietf.org/html/rfc826
+///
+// NOTE: Could be implemented in various ways, such as a specialized version of EthernetFrame that's
+// known to be an ARP frame. It could be implemented in a similar way that packets are promoted/demoted
+// with `TryFrom`.
+#[derive(Clone)]
+pub struct ArpFrame {
+    frame: EthernetFrame,
```
CollinValley, line 26: I think this is causing you to fight an uphill battle to manipulate the EthernetFrame data directly. Why not just make a new primitive Packet type,  following the same pattern EthernetFrame, IPv4, IPv6, etc follow. Rather than wrap an EthernetFrame, just take over it's PacketData vec.

CollinValley: You can then implement try from following the same pattern as others.

route-rs-packets/src/arp.rs
```diff
@@ -0,0 +1,253 @@
...
+    Request = 1,
+    Reply = 2,
+}
+
+pub enum ArpHardwareType {
+    Ethernet = 1,
+}
+
+// TODO: should move to types.rs or ethernet.rs
+pub const IPV4_ETHER_TYPE: u16 = 0x0800;
```
CollinValley: Sure, shared constants and Types belong in Types.rs

route-rs-packets/src/arp.rs
```diff
@@ -0,0 +1,253 @@
...
+// known to be an ARP frame. It could be implemented in a similar way that packets are promoted/demoted
+// with `TryFrom`.
+#[derive(Clone)]
+pub struct ArpFrame {
+    frame: EthernetFrame,
+}
+
+impl ArpFrame {
+    // TODO: would like default args for Ethernet (6 bytes) and IPv4 (4 bytes)
+    // TODO: Should we accept a struct of options to initialize the ArpFrame with? Or use a builder pattern?
```
CollinValley: So far we have done neither in Packets, since most frames are not really initialized from scratch. They are "imported" so to speak, off the NIC from a Vec<u8>.  In Packets we have empty() and from_buffer() as the interface functions to make new Packets.  In from_buffer() just do some basic sanity checking that the vector provided is an ARP frame, (think correct EtherType and Len(), not much else).  empty() should just do what you expect, but you can at least provide a performance improvement by pre-allocating the data vec! to be the correct size, and set the EtherType.

route-rs-packets/src/arp.rs
```diff
@@ -0,0 +1,253 @@
...
+// NOTE: Could be implemented in various ways, such as a specialized version of EthernetFrame that's
+// known to be an ARP frame. It could be implemented in a similar way that packets are promoted/demoted
+// with `TryFrom`.
+#[derive(Clone)]
+pub struct ArpFrame {
+    frame: EthernetFrame,
+}
+
+impl ArpFrame {
+    // TODO: would like default args for Ethernet (6 bytes) and IPv4 (4 bytes)
```
CollinValley: Just default to 0s,  this is a "thin" library in that there are not many guardrails, it will create invalid packets if asked.  It is up to the user to set all values other than those that are immutable to the type (such as correct EtherType for an ARPFrame).

route-rs-packets/src/arp.rs
```diff
@@ -0,0 +1,253 @@
...
+        let mut frame = EthernetFrame::empty();
+        frame.set_payload(payload.as_slice());
+
+        let mut arp_frame = ArpFrame { frame };
+        arp_frame.set_hardware_addr_len(hardware_addr_len);
+        arp_frame.set_protocol_addr_len(protocol_addr_len);
+        arp_frame
+    }
+
+    pub fn from(frame: EthernetFrame) -> Self {
```
CollinValley: Let's not reimplement from, but not in the From trait, that is confusing.  Also I suggest TryFrom instead of From, since not all EthernetFrame's are ArpFrame's and we should reflect that in our interface.

route-rs-packets/src/arp.rs
```diff
@@ -0,0 +1,253 @@
...
+    pub fn frame(self) -> EthernetFrame {
+        self.frame
+    }
+
+    // Returns the bytes in the ethernet frame between start and end, exclusive
+    fn arp_data(&self, start: usize, end: usize) -> &[u8] {
+        let frame_offset_start = self.frame.payload_offset + start;
+        let frame_offset_end = self.frame.payload_offset + end;
+
+        // TODO: I'd like to use `self.frame.payload()` here, but having ownership difficulties with Cow
```
CollinValley, line 151: Yeah, this is the battle that you don't "own" the vector, you "own" the EthernetFrame.  So instead of containing the EthernetFrame, just strip the PacketData out into you structure, and set the layer2_offset appropriately.

route-rs-packets/src/arp.rs
CollinValley: Struct should have its own PacketData, and offset pointers.

CollinValley (changes requested):
I think this is a good baseline, but the implementation you've chosen isn't really in line with the implementations of the rest of the library.  I think if you mirrored those implementations your ownership issues would resolve themselves.

route-rs-packets/src/arp.rs
ddowl: I actually disagree a bit here. I like that we already have an abstraction for dealing with header offsets and underlying data of a packet. `ArpFrame` is only supposed to encapsulate the details of the _payload_ of an ARP frame, but by unwrapping `EthernetFrame` we take on the additional complexity of knowing about the underlying buffer and necessary offsets. I'd rather hide that complexity inside of `EthernetFrame`.

route-rs-packets/src/arp.rs
ddowl: Cool. I think `new` fits that role of `empty`. I don't think there are traits defining this initialization style, so we'll just hold it as convention?

I'm also considering the only case ARP frames get created, the "Packet Generation" section of the [RFC](https://tools.ietf.org/html/rfc826), but we can keep this as a thin wrapper to stay in line with the rest of the lib.

route-rs-packets/src/arp.rs
ddowl: I guess another way to think about it is that `ArpFrame` isn't a primitive packet construct. An `ArpFrame` _is_ an `EthernetFrame` with additional constraints and ways to view its payload, so it seems like it should be composed of an `EthernetFrame` and use it as a client.

route-rs-packets/src/arp.rs
ddowl: I need to set an empty ARP frame to have _some_ values here, since the payload is variable sized. We run into some subtle bugs if we don't set these values to something reasonable, e.g. If we attempt to set any of the variable length payload fields while these lengths are 0, we'll have nowhere to write to.

Since we're wrapping `EthernetFrame` here, I think it's ok to provide a few more guardrails.

route-rs-packets/src/arp.rs
ddowl: Good idea, I didn't think of that. Should we support both

`try_from(frame: EthernetFrame)` and `from_buffer(buf: PacketData)`?

route-rs-packets/src/arp.rs
ddowl: Yeah, I see the lifetime requirements now. If I own the packet vector (as a field in the `ArpFrame` struct) we know any references that are passed out here will live through the caller (at least until `ArpFrame` is `Drop`ped?).

ddowl: I guess we're still using `EthernetFrame` as a client effectively here, it just seems like exposing too much of its internals. As `ArpFrame`, I shouldn't have to know about the internal structure of `EthernetFrame` in order to obtain a reference to its payload.

ddowl: This is more of a nit point I'd like to make people aware of than an urgent problem.

CollinValley: Hmm, the PacketData field is public, so as an owner of EthernetFrame you are able to directly change its vec, you just do so at your own peril. 
Payload is supposed to a be a take it and go kind of function.  Or you can wholesale take the payload, modify it, and drop it back into set_payload().  Does cause an allocation though.

route-rs-packets/src/arp.rs
CollinValley: Yeah, typically, I've just made `try_from()` call `from_buffer()` and left all the validation logic inside of `from_buffer`.  If you decide to keep wrapping an EthernetFrame rather than have the raw vector, you may need something slightly different.

route-rs-packets/src/arp.rs
ddowl: Exactly! I'd like to only get/set particular ranges of bytes in the underlying vec, but I'm finding it difficult to do that with `EthernetFrame`'s interface atm. Again, not the biggest deal since we're able to reference its `PacketData` here. I'm also able to modify it in the `set_arp_data` fn.

route-rs-packets/src/arp.rs
CollinValley: Yeah no traits right now, maybe they should be pulled into the `Packet` trait, but I've left it flexible for now.  Also, feel free to leave more bespoke convenience constructors that take more parameters, if you think it is appropriate.

ddowl: For sure. I might do that when using this for ARP frame generation in a future PR.

route-rs-packets/src/arp.rs
CollinValley: Typically I've handled Length fields by setting them appropriately compared to the length of the buffer we make.  I've also noticed the protocols usually specify a valid minimum value.  Does something like that apply here?

route-rs-packets/src/arp.rs
CollinValley: Okay, using the `PacketData` escape hatch is fine for now.  Rust's safety might be getting in the way of giving us the interface we would like trivially.

route-rs-packets/src/arp.rs
CollinValley: I guess the short answer here, is pick some reasonable defaults, and document it in the doc comments.

ddowl: Maybe? I could try to find the minimum hardware/protocol addr lengths that ARP supports, but those parameters would only work with that particular hardware/protocol pair. I think I'll brainstorm some better constructor ideas to make sure the packet gets initialized into a consistent state.

route-rs-packets/src/arp.rs
```diff
@@ -0,0 +1,253 @@
...
+        self.set_ip_addr(ip_addr, start);
+    }
+
+    // Move ownership of the frame back to the caller
+    pub fn frame(self) -> EthernetFrame {
+        self.frame
+    }
+
+    // Returns the bytes in the ethernet frame between start and end, exclusive
+    fn arp_data(&self, start: usize, end: usize) -> &[u8] {
```
CollinValley, line 147: You are returning a slice here, which has to have lifetimes. You could also return a COW, then you sidestep the lifetime issues, and may be able to call EthernetFrame's payload() function.

route-rs-packets/src/arp.rs
CollinValley: Hmm, if you want to continue to wrap the EthernetFrame, that is fine, but it breaks from the way the rest of the Packets are laid out.  Also, tbf, I don't see many spots where you are you are actually using the EthernetFrame outside of reaching through it to the PacketData, or getting the payload/set_payload().  I'm not sure you are actually minimizing complexity by not stripping the PacketData and offsets into your own structure.  You also can't support the inside-out construction (encap functions) that other higher level Packets do, where you make a bare TCPHeader first, then encap it with IPv4, so on and so forth.  But if you think that the wrapped frame is the right way to go I trust your judgement since you have a better handle on how this is going to be used in the field than I do.

route-rs-packets/src/arp.rs
```diff
@@ -0,0 +1,275 @@
...
+        let (start, end) = self.target_hardware_addr_range();
+        self.arp_data(start, end)
+    }
+
+    pub fn target_protocol_addr(&self) -> &[u8] {
+        let (start, end) = self.target_protocol_addr_range();
+        self.arp_data(start, end)
+    }
+
+    pub fn set_hardware_type(&mut self, htype: u16) {
```
CollinValley: Last change, this one is a new library thing, but trivial to add. 
Please make every setter function return a `&mut Self` , this lets the user optionally chain setters. 
Like
A::new().set_1().set_2().set_3();

And the old uses are still valid since Rust will auto drop  returned references that are not assigned to a variable. See the other packets files for examples.

CollinValley (commented):
Last thing, after this is ready for merge.

ddowl:
@CollinValley Just pushed up the 2 relevant fixes we talked about in the last meeting:
1. Chaining setters with `&mut Self`
2. Failing `TryFrom::<EthernetFrame>` tests

Interestingly, I found that we can't actually chain setters from the initial creation of an `ArpFrame`, like 

```
let arp_frame = ArpFrame::new().set_a(a).set_b(b).set_c(c);
```

since `ArpFrame::new()` returns `Self` (the caller owns the newly created wrapped `EthernetFrame`). The first setter call will actually return a borrow of the frame, so the caller never actually receives ownership before modification. This results in the ownership being "dropped while borrowed".

Screen Shot 2020-03-21 at 4 29 42 PM

However, if the caller actually receives full ownership of the frame, then we can string our setters together:

```
let mut arp_frame = ArpFrame::new();
arp_frame
    .set_a(a)
    .set_b(b)
    .set_c(c);
```

CollinValley (approved):
LGTMMD

PR #214, opened 2020-03-08
travis.yml: Fix missing libpcap

scgruber:
The pcap crate requires libpcap, so let's make sure that's installed in
our travis setup.


CollinValley:
The check fails but that is expected.

PR #215, opened 2020-03-09
Draft: AutoEgressor, auto unpack egressor vector

CollinValley:
This is a super trivial commit using a macro to auto unpack a the
egressor vector into a generic length set of variables. I bet with some
more clever thinking we could make something even more ergonomic, but I
wanted to at least demonstrate this is possible. The macro is quite
simple, checkout rust_by_example variadic functions for more.

The macro basically takes the name of the vector, and an unlimited length of identifiers to pull the values of the vector into. Under the hood it simply injects the remove() call, in a recursive manner. 

I don't think this one is ready for merge, maybe in person we can brainstorm what a better layout would look like, but this is pretty good for a 10 min whip up.

### DIFF
--- a/examples/minimal-static-router/src/main.rs
+++ b/examples/minimal-static-router/src/main.rs
@@ -4,6 +4,7 @@ use route_rs_runtime::link::{
     Link, LinkBuilder, PacketStream, ProcessLinkBuilder,
 };
 use route_rs_runtime::utils::{runner::runner, test::packet_generators::immediate_stream};
+use route_rs_runtime::AutoEgressor;
 
 mod classifiers;
 mod processors;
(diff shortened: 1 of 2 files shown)


examples/minimal-static-router/src/main.rs
```diff
@@ -160,19 +161,20 @@ impl LinkBuilder<EthernetFrame, EthernetFrame> for Router {
                     }))
                     .build_link();
             all_runnables.append(&mut ipv6_subnet_router_runnables);
+            AutoEgressor!(ipv6_subnet_router_egressors, sub_a, sub_b, sub_c);
```
ddowl, line 164: It's cool that this macro can extract each of the `Vec`'s values into different variables. The only nits I have are 

1. The macro itself doesn't have anything to do with egressors. It just unpacks `Vec`s. Maybe `vecUnpack!`?
2. If you didn't supply the right number of values we would get weird behavior, like trying to remove from an empty `Vec` I imagine. This version will probably panic at runtime while the router is being initialized, but it would be nice if we could verify the number of variables matches the length of the vec earlier.

ddowl:
Whoops, sorry.

PR #217, opened 2020-03-13
Unpack!: A Macro to help you remove stuff from links

CollinValley:
This is a preliminary cut, for a draft commit
It does two things at once:
For runnables, it takes the vector passed in, your `all_runnables` and
appends the runnables returned from the link to the provided vector. You
may also choose to provide a `_` instead, and the runnables are
discarded.

For egressors, if you provide a single variable, it is declared as you
would expect, and represents the vector of egressors that is returned
from the link. If you provide a one or more variables inside of `[]`
square brackets, it automatically destructures the vector into those the
provided variables.

Examples
unpack!( let (runnables, egressors) = myLink::new().build; );

unpack!( let (runnables, [a, b, c]) = myLink::new().build; );

unpack!( let (_, egressors) = myLink::new().build; );

unpack!( let (_, [a,b,c]) = myLink::new().build; );

### DIFF
--- a/examples/pi-home-router/src/interface/link/interface_collect.rs
+++ b/examples/pi-home-router/src/interface/link/interface_collect.rs
@@ -5,6 +5,7 @@ use route_rs_runtime::link::{
     primitive::{JoinLink, ProcessLink},
     Link, LinkBuilder, PacketStream, ProcessLinkBuilder,
 };
+use route_rs_runtime::unpack;
 
 /// InterfaceCollect: Link that our 3 input interfaces, Host, Lan and Wan, expected in that order.
 pub(crate) struct InterfaceCollect {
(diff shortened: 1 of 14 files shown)


CollinValley:
Whoops, this is supposed to be more of a draft commit. RFC type thing.  Do take a look though

ddowl (approved):
Neat stuff! You could have the macro change target `master` if you like. We can also change this up as more people get a feel for what good ergonomics feel like, but I don't mind merging this in if we use it sparingly.

CollinValley:
Thanks, Drew!  I'm going to hold off for a sec, waiting for @scgruber to weigh in, this is primarily to help his efforts anyways. Will consider jumping this to master too.

scgruber (approved):
This seems like a good approach. My only complaint is that IntelliJ doesn't currently do any analysis on content inside of macros, so this kind of trashes the code sense. Hopefully the devs will improve this in the future.

PR #218, opened 2020-03-15
Runtime: Primitive: Switch to default unbounded channels

CollinValley:
Switch to using unbounded channels inside our primitive channel using
links (Queue, Join, Fork and Classify), if the user does not specify a
channel size. Should they specify a size through the same interface, we
provide them a bounded channel.

This is a good change because the kernel (through the AFPacket
interface) originally allocates the memory for us, we just pass fat
pointers around through the vectors the represent the data in our
packets. Because of this fact, adding and removing things to channels
does not cause an allocation, so it the RAM usage does not really grow
as more and more packets are "added".  Dropped packets due to overflow 
would (usually) only happen in the
event that the packet rate exceeds the line rate of our NIC.  The total
number of packets "in-flight" in the router, and hence the total RAM
being used, will be bounded by this ring-buffer exposed to user space.
So users of the runtime shouldn't really have to worry about forcing
various buffer points to be a certain size, the RAM usage off the
program won't grow out of control. This is one of the powers of relying
on Rust, we have good control over the allocations made and can keep the
runtime fast.

Because we are multithreaded, concurrent, and even semi-parallel (multi-cpu), we
should try to let tasks do as much work as possible before they exit,
since this keeps the I-Cache on the CPU hotter. If we set arbitrary
channel size limits, we force the task to switch off before it has
exhausted its input pipe, effectively flushing our cache.

### DIFF
--- a/route-rs-runtime/src/link/primitive/classify_link.rs
+++ b/route-rs-runtime/src/link/primitive/classify_link.rs
@@ -17,7 +17,7 @@ pub struct ClassifyLink<C: Classifier> {
     in_stream: Option<PacketStream<C::Packet>>,
     classifier: Option<C>,
     dispatcher: Option<Box<Dispatcher<'static, C::Class>>>,
-    queue_capacity: usize,
+    queue_capacity: Option<usize>,
     num_egressors: Option<usize>,
 }
 
@@ -27,7 +27,7 @@ impl<C: Classifier> ClassifyLink<C> {
             in_stream: None,
             classifier: None,
             dispatcher: None,
-            queue_capacity: 10,
+            queue_capacity: None,
             num_egressors: None,
         }
     }
@@ -47,7 +47,7 @@ impl<C: Classifier> ClassifyLink<C> {
             queue_capacity > 0,
             format!("Queue capacity: {}, must be > 0", queue_capacity)
         );
-        self.queue_capacity = queue_capacity;
+        self.queue_capacity = Some(queue_capacity);
         self
     }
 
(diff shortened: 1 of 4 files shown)


route-rs-runtime/src/link/primitive/classify_link.rs
```diff
@@ -104,8 +104,20 @@ impl<C: Classifier + Send + 'static> LinkBuilder<C::Packet, C::Packet> for Class
...
+                        let (s, r) = crossbeam_channel::unbounded::<Option<C::Packet>>();
+                        to_egressor = s;
+                        from_ingressor = r;
+                    }
+                    Some(capacity) => {
+                        let (s, r) = crossbeam_channel::bounded::<Option<C::Packet>>(capacity);
+                        to_egressor = s;
+                        from_ingressor = r;
+                    }
+                }
```
ddowl: I think this block can be more concisely expressed as:
```
let (to_egressor, from_ingressor) = match self.queue_capacity {
  None => crossbeam_channel::unbounded::<Option<C::Packet>>(),
  Some(capacity) => crossbeam_channel::bounded::<Option<C::Packet>>(capacity)
};
```

CollinValley: /facepalm Yeah, good point.

ddowl: Since this section seems to be repeated for each link, maybe we can make a runtime utility function for `internalLinkChannel(capacity: Option<usize>)` or something that each link can share.

ddowl (commented):
Conceptually, the change seems fine to me. Users don't need to worry about their unbounded memory growth since we're already passing around pointers that are allocated, and packet dropping will occur at the NIC's level.

I'm still concerned that users could potentially create a graph that backs up _locally_ in certain areas. It makes sense to optionally restrict queue size (want to utilize that cache), but we should also be able to report "unbounded growth" of internal queues upon request. Maybe this could be a future feature we can make an issue for.

route-rs-runtime/src/link/primitive/classify_link.rs
CollinValley: Hmm, I'm not sure I agree, it is only 3 lines of code, and the function will have to be generic over the type we pass around inside the link, might be more trouble that its worth for four locations.

ddowl (approved):
I don't think we'll run into those kinds of problems soon though, so this LGTMFNMD. Thanks for the quick update!

PR #224, opened 2020-03-31
PHR: ArpHandler & ArpGenerator

ddowl:
Introduces 2 new Links to PHR, `ArpHandler` and `ArpGenerator` (not tied to those names). They implement handling and generating ARP requests respectively, using `route-rs-packets::ArpFrame` and a shared `arp_table: Arc<Mutex<HashMap<Ipv4Addr, MacAddr>>>`. There are plenty of things that can be improved, but I believe this implementation is functional with a minimal test suite. Feel free to skim the code for `TODO`s and leave your 2 cents.

### DIFF
--- a/examples/pi-home-router/src/arp/arp_generator.rs
+++ b/examples/pi-home-router/src/arp/arp_generator.rs
@@ -1,0 +1,184 @@
+use crate::types::InterfaceAnnotated;
+use route_rs_packets::{
+    ArpFrame, ArpOp, EthernetFrame, Ipv4Packet, MacAddr, ARP_ETHER_TYPE, IPV4_ETHER_TYPE,
+};
+use route_rs_runtime::processor::Processor;
+use std::collections::HashMap;
+use std::convert::TryFrom;
+use std::net::{IpAddr, Ipv4Addr};
+use std::sync::{Arc, Mutex};
+
+// TODO: This could also be called AddMacOrGenArp
+pub(crate) struct ArpGenerator {
+    arp_table: Arc<Mutex<HashMap<Ipv4Addr, MacAddr>>>,
+}
+
+impl ArpGenerator {
+    pub fn new(arp_table: Arc<Mutex<HashMap<Ipv4Addr, MacAddr>>>) -> Self {
+        ArpGenerator { arp_table }
+    }
+}
+
+impl Processor for ArpGenerator {
+    type Input = InterfaceAnnotated<EthernetFrame>;
+    type Output = InterfaceAnnotated<EthernetFrame>;
+
+    ///
+    /// From the ARP RFC: https://tools.ietf.org/html/rfc826
+    ///
+    /// As a packet is sent down through the network layers, routing determines the protocol address
+    /// of the next hop for the packet and on which piece of hardware it expects to find the station
+    /// with the immediate target protocol address. The Address Resolution module converts the
+    /// <protocol type, target protocol address> pair to a 48.bit Ethernet address. The Address
+    /// Resolution module tries to find this pair in a table. If it finds the pair, it gives the
+    /// corresponding 48.bit Ethernet address back to the caller which then transmits the packet.
+    ///
+    /// If the lookup for the (protocol type, protocol address) fails, it probably informs the
+    /// caller that it is throwing the packet away (on the assumption the packet will be
+    /// retransmitted by a higher network layer), and generates an Ethernet packet with a type field
+    /// of ether_type$ADDRESS_RESOLUTION. The Address Resolution module then sets
+    ///
+    /// Hardware Type: ares_hrd$Ethernet = 1
+    /// Protocol Type: type that is being resolved
+    /// Hardware Address Length: 6 (the number of bytes in a 48.bit Ethernet address)
+    /// Protocol Address Length: length of an address in that protocol
+    /// Op: ares_op$REQUEST = 1
+    /// Sender Hardware Address: the 48.bit ethernet address of itself
+    /// Sender Protocol Address: the protocol address of itself
+    /// Target Protocol Address: the protocol address of the machine that is trying to be accessed.
+    ///
+    /// It does not set Target Hardware Address to anything in particular, because it is this value
+    /// that it is trying to determine. It could set the Target Hardware Address to the broadcast
+    /// address (FF:FF:FF:FF:FF:FF) for the hardware (all ones in the case of the 10Mbit Ethernet)
+    /// if that makes it convenient for some aspect of the implementation.
+    ///
+    /// It then causes this packet to be broadcast to all stations on the Ethernet cable originally
+    /// determined by the routing mechanism.
+    ///
+    fn process(&mut self, packet: Self::Input) -> Option<Self::Output> {
+        let mut frame = packet.packet;
+        match frame.ether_type() {
+            IPV4_ETHER_TYPE => {
+                let ipv4_packet = Ipv4Packet::try_from(frame).unwrap();
+
+                let ipv4_src_addr = ipv4_packet.src_addr();
+                let ipv4_dest_addr = ipv4_packet.dest_addr();
+                frame = EthernetFrame::encap_ipv4(ipv4_packet);
+
+                match self.arp_table.lock().unwrap().get(&ipv4_dest_addr) {
+                    Some(&dest_mac_addr) => {
+                        // We found a MacAddr for this (protocol type, target address) tuple!
+                        // Let's set it, and pass the packet along
+                        frame.set_dest_mac(dest_mac_addr);
+                        Some(InterfaceAnnotated::<EthernetFrame> {
+                            packet: frame,
+                            inbound_interface: packet.inbound_interface,
+                            outbound_interface: packet.outbound_interface,
+                        })
+                    }
+                    None => {
+                        // Couldn't find a match in the table, let's generate an ARP request
+                        let mut arp_request = ArpFrame::default();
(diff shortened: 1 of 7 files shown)


examples/pi-home-router/src/arp/arp_handler.rs
```diff
@@ -0,0 +1,212 @@
...
+    ///             ?Is the opcode ares_op$REQUEST?  (NOW look at the opcode!!)
+    ///             Yes:
+    ///                 Swap hardware and protocol fields, putting the local hardware and protocol
+    ///                     addresses in the sender fields.
+    ///                 Set the ar$op field to ares_op$REPLY
+    ///                 Send the packet to the (new) target hardware address on the same hardware on
+    ///                     which the request was received.
+    ///
+    fn process(&mut self, packet: Self::Input) -> Option<Self::Output> {
+        let mut arp_frame = ArpFrame::try_from(packet.packet).unwrap();
```
ddowl, line 54: There's a few different options for how we can handle invalid ARP frames. We can either
1. Do validation outside of this Processor so `Self::Input == ArpFrame` (a little tighter bounds)
2. Validate here and drop on error
3. Validate here, log error, then drop
4. Validate here, forward to logging element which drops.

examples/pi-home-router/src/arp/arp_handler.rs
```diff
@@ -0,0 +1,212 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::types::Interface;
+    use route_rs_packets::ARP_ETHER_TYPE;
+
+    fn stub_annotated_input(frame: EthernetFrame) -> InterfaceAnnotated<EthernetFrame> {
+        InterfaceAnnotated::<EthernetFrame> {
```
ddowl, line 119: Attempt to remove turbofish here.

examples/pi-home-router/src/arp/arp_handler.rs
```diff
@@ -0,0 +1,212 @@
...
+                            .insert(sender_ipv4_address, sender_mac_address);
+                    }
+
+                    if arp_frame.opcode() == ArpOp::Request as u16 {
+                        // Generate response ARP frame
+                        arp_frame.set_target_hardware_addr(sender_mac_address);
+                        arp_frame.set_target_protocol_addr(IpAddr::V4(sender_ipv4_address));
+                        // TODO: how can I find this machine's MacAddr?
+                        // Will it be pre-populated in the ARP table?
+                        arp_frame.set_sender_hardware_addr(MacAddr::new([0, 0, 0, 0, 0, 0]));
```
ddowl, line 91: Assume that this machine's MAC and IPv4 addresses will be passed in on initialization.

route-rs-packets/src/arp.rs
```diff
@@ -143,6 +143,25 @@ impl ArpFrame {
         self.frame
     }
 
+    /// Default-specific methods
+    pub fn sender_mac_addr(&self) -> Result<MacAddr, std::array::TryFromSliceError> {
```
ddowl, line 147: Let's do `use std::array::TryFromSliceError` so we don't need to fully qualify it each time.

examples/pi-home-router/src/arp/arp_generator.rs
```diff
@@ -0,0 +1,184 @@
...
+                        Some(InterfaceAnnotated::<EthernetFrame> {
+                            packet: frame,
+                            inbound_interface: packet.inbound_interface,
+                            outbound_interface: packet.outbound_interface,
+                        })
+                    }
+                }
+            }
+            // Ignore non-IPv4 packets
+            // TODO: add error logging?
```
CollinValley, line 102: Why not just return None here to indicate something incorrect?

examples/pi-home-router/src/arp/arp_generator.rs
```diff
@@ -0,0 +1,184 @@
...
+    /// if that makes it convenient for some aspect of the implementation.
+    ///
+    /// It then causes this packet to be broadcast to all stations on the Ethernet cable originally
+    /// determined by the routing mechanism.
+    ///
+    fn process(&mut self, packet: Self::Input) -> Option<Self::Output> {
+        let mut frame = packet.packet;
+        match frame.ether_type() {
+            IPV4_ETHER_TYPE => {
+                let ipv4_packet = Ipv4Packet::try_from(frame).unwrap();
```
CollinValley, line 62: We should handle the unwrap() here. Why not return None if the try_from fails?

examples/pi-home-router/src/arp/arp_handler.rs
CollinValley: Can we use the type system here to provide us an `IntAnt<ArpFrame>`. Seems closest to what we wanted this system to do in the first place, use the static types to prevent errors.

CollinValley: My second choice is to drop on error, ignore logging for now.

examples/pi-home-router/src/arp/arp_generator.rs
CollinValley: optionally, could we use a stricter input type to solve this issue?

examples/pi-home-router/src/arp/arp_handler.rs
CollinValley: You should take the required values as arguments in your new function.  Make sure that the router won't build without the required init variables.  This info will have to be passed into the router, probably via config file, during startup.  I wager rust has a good .toml parser or something lol.

examples/pi-home-router/src/router.rs
```diff
@@ -1,11 +1,27 @@
...
     wan: Option<String>,
+
+    // TODO: investigate/discuss concurrency options
+    // https://doc.rust-lang.org/std/sync/struct.RwLock.html
+    // https://docs.rs/chashmap/2.2.2/chashmap/
+    // https://github.com/jonhoo/rust-evmap
+    // https://github.com/jonhoo/flurry
+    /// Top-level ARP table reference.
+    /// Links that require a reference should clone this Arc.
+    arp_table: Arc<Mutex<HashMap<Ipv4Addr, MacAddr>>>,
```
CollinValley, line 24: Another option https://github.com/xacrimon/dashmap.  Its concurrent out of the box, so perhaps this will be easier to use.

examples/pi-home-router/src/router.rs
```diff
@@ -55,6 +75,19 @@ impl LinkBuilder<(), ()> for Router {
...
         // all the relevant runnables, and no egressors(Since our router has none)
+
+        // TODO: wire up from protocol classifier
+        let _arp_handler = ProcessLink::new()
+            .ingressor(immediate_stream(vec![]))
+            .processor(ArpHandler::new(self.arp_table.clone()))
+            .build_link();
+
+        // TODO: not sure where to put ArpGenerator
+        let _arp_generator = ProcessLink::new()
```
CollinValley, line 86: When we draw the big ol graph we will know where to put these two.

route-rs-packets/src/arp.rs
CollinValley: Is it necessary to return a result here?  Will a valid ArpFrame ever return values that would cause this to fail?
For instance, will self.sender_hardware_addr() return a slice of values that would cause MacAddr to not be able to make an MacAddr?  The only case this happens is if the slice is too short, which won't ever happen unless the underlying vector has been emptied, which isn't a case the library actually handles, as it exposes the array publicly.  

I think you should  follow the rest of the lib and return the actual value and panic if you get a slice error. (see Ipv4Packet.src_addr().  This reduces the load on our end user significantly.

CollinValley: if we want to never panic, we should probably do that all at once in a separate PR.  For now I'd like to keep the API of the library consistent across each type.

route-rs-packets/src/arp.rs
```diff
@@ -199,6 +221,22 @@ impl ArpFrame {
         let end = start + plen;
         (start, end)
     }
+
+    fn bytes_func_to_ipv4(&self, bytes: &[u8]) -> Result<Ipv4Addr, std::array::TryFromSliceError> {
```
CollinValley, line 225: Same as point as the comments above.

examples/pi-home-router/src/arp/arp_handler.rs
```diff
@@ -0,0 +1,212 @@
...
+
+        // Addressed to router, so ARP frame is a Reply, with updated sender values
+        assert_eq!(output.packet.ether_type(), ARP_ETHER_TYPE);
+        let output_arp_frame = ArpFrame::try_from(output.packet).unwrap();
+        assert_eq!(output_arp_frame.opcode(), ArpOp::Reply as u16);
+        assert_eq!(
+            output_arp_frame.sender_ipv4_addr().unwrap(),
+            ROUTER_IPV4_ADDR
+        );
+        // assert_eq!(
```
CollinValley, line 202: Nit: Delete this extraneous case.

route-rs-packets/src/arp.rs
CollinValley: As far as I can tell, your arp_data function would actually panic during accessing the vec via the index operator.  That won't get caught because it returns a &[u8],  so the slice out of bounds error causes a panic.  Therefore, the above `Result`s will never return an error.

CollinValley (commented):
Some stuff to consider.  I think the Arp stuff looks really solid. Many thanks

